PR #5, opened 2015-07-23
Story/quad 2

jdamerow:
Replaced CollaboratorRole with QuadrigaRole as they were basically identical. Some bugfixes. Fifth time for Jenkins.


jdamerow:
ok to test

PR #9, opened 2015-09-24
Story/quad 14

parag2507:
Changes for deactivate workspace


jdamerow:
ok to test

jdamerow:
See comments.

PR #16, opened 2015-09-29
Story quad 5

mkarthik90:
Fixed the alignment issues by using indent. Removed success message and used alert box.


jdamerow:
let's wait till all subtasks are done

PR #25, opened 2015-10-07
Story/quad 16

HariChandanaKanchanapally:
Added roles for wokspace and merged the fetch user name line


jdamerow:
ok to test

jdamerow:
only activate or deactivate workspace should be enabled, depending on what action the user can perform: if the workspace is activated "deactivate" should be enabled, otherwise "activate"
image

PR #26, opened 2015-10-07
story/QUAD-27

HariChandanaKanchanapally:
Inactive Workspace page lists decativated workspaces like activated workspace list, Added access permission for viewing deactivated list to project roles


jdamerow:
ok to test

PR #32, opened 2015-10-09
Story/quad 16

HariChandanaKanchanapally:
Redirected to activated workspace instead of navigating to project page on activation


jdamerow:
ok to test

jdamerow:
Two things:
1) please redirect back to the workspace after deactivating/activating it, not to the project
2) the alert dialog should be further up, so that I don't have to scroll to click ok. Also, it should show a message like "Are you sure you want to deactivate the workspace?" and then yes or no as buttons.

HariChandanaKanchanapally:
Sure. I have changed the redirect for activation. I'll change it for
deactivation too. Also, about the alert box, I'll make the changes during
the next class.

Regards,
Chandana

PR #38, opened 2015-10-14
Story quad7

parag2507:
Pull request after addressing review comments


jdamerow:
ok to test

Quadriga/src/main/webapp/WEB-INF/views/workbench/project.jsp
```diff
@@ -141,8 +141,41 @@
...
+											"No" : function() {
+												$(this).dialog('close');
+												return false;
+											}
+										}
+									});
+				}
+			</script>
+			<div id="dialog-delete-project-confirm" title="Confirm Delete?">
+				You are about to delete a project, this is not reversible.</br> Do you want to proceed?.</div>
```
jdamerow, line 178: There should be no period after "Do you want to proceed?".
Also, right now the dialog opens at the bottom of the page so that is it out of view when I click on "Delete project" (see screenshot). Please make it open further up, so that the user doesn't have to scroll first.
image

jdamerow:
see comment

PR #39, opened 2015-10-14
Story quad 5

mkarthik90:
Fixed Review comments


jdamerow:
ok to test

jdamerow:
For edit project url:
1) I shouldn't be able to edit the project name (only custom url -> can we rename that to project url)
2) I can submit an empty custom url field (shouldn't be possible)
3) Can we add a warning message for updating the url that says that if the user changes the project url, links to the old url won't work anymore?

image

jdamerow:
see comment

PR #46, opened 2015-10-16
Story quad 5

mkarthik90:
Resolved merge conflicts


jdamerow:
ups, did you format everything? Try to avoid that in the future, it makes it impossible to review the code because it's too many files. It's not a big deal right now, since I believe I saw everything already, but for next time.

jdamerow:
ok to test

jdamerow:
I'll merge this branch, but please change the following in your next commit. 
1. Instead of 
"Old URL will not work anymore. Do you wish to continue?"
it should be 
"Please be aware that changing the custom URL will disabled the old project URL. Do you wish to continue?"
1. The validation method still says "Unix name should be unique." This should be changed to "Custom URL has to be unique."

Also, right now, if I try to submit the form without changing the custom URL, it shows a validation error saying that the unix name has to be unique. This doesn't make much sense, since it is still unique (I'm not changing it). We should probably only enable the submit button until the custom url was changed to avoid that. But I'll create a new ticket for that.

PR #49, opened 2015-10-26
Story/quad 16

HariChandanaKanchanapally:
Changes made:

The Confirm alert boxes appear in the middle of the page, so one doesn't have to drag down to confirm any activity on workspace
The deactivateworkspace redirects to the deactivated workspace instead of project
The delete workspace previously deleted dictionary and then updated the DTO. This caused deleted instance passed to update(). Did not delete the dictionaries because they have to remain even after the workspace is deleted


jdamerow:
ok to test

jdamerow:
Please put the question in the dialog into the content field, and have the header say something like "Activate Workspace".

jdamerow:
Also, please make the same change for deactivating a workspace (Header: Deactivate Workspace, Body: the question). Sorry, I oversaw that when checking the other story. Thanks!

jdamerow:
Also, can you please remove these two buttons from the deactivated workspace page:
image

PR #50, opened 2015-10-28
Story/quad 16

HariChandanaKanchanapally:
1. Header: Deactivate/Activate/Delete Workspace, Body: the question
2. Removed the Add text from Dspace, Change Dspace Login from deactivated workspace


jdamerow:
ok to test

PR #51, opened 2015-11-04
Story/quad 38

HariChandanaKanchanapally:
Allow passthrough projects without checking if the DSpace is present in the XML


Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkManager.java
```diff
@@ -1191,12 +1192,12 @@ public String storeNetworkDetails(String xml, IUser user, String networkName,Str
...
 		newNetworkDetailCache = parseNewNetworkStatement(elementEventType,workspaceBitStreamList,newNetworkDetailCache);
 
-		// Check if it DSpace is present in the XML
+		/*// Check if it DSpace is present in the XML
 		if(!newNetworkDetailCache.isFileExists()){
 			logger.info("Network not uploaded");
 			logger.info("Some of the text files in the uploaded network were not present in the workspace");
 			return INetworkManager.DSPACEERROR;
-		}
+		}*/
```
jdamerow, line 1200: please just remove the code instead of commenting it out.

PR #53, opened 2015-11-06
Test/quad 45

suprajkolluri:
All the classes mentioned above already have test cases under the package edu.asu.spring.quadriga.service.impl.profile. Moved them to the correct package.
Also ViafManager1 is not being used anywhere. Deleting it.


jdamerow:
ok to test

jdamerow:
ups, cool :)

PR #55, opened 2015-12-09
Test/quad 42

suprajkolluri:
Refactored the code as per review comments.
Added new dependency jmockit to mock string class as final classes cannot be mocked using mockito. Also formatted the code correctly.

### DIFF
--- a/Quadriga/src/test/java/edu/asu/spring/quadriga/service/impl/conceptcollection/ConceptCollectionManagerTest.java
+++ b/Quadriga/src/test/java/edu/asu/spring/quadriga/service/impl/conceptcollection/ConceptCollectionManagerTest.java
@@ -35,175 +35,185 @@ import edu.asu.spring.quadriga.exceptions.QuadrigaStorageException;
 import edu.asu.spring.quadriga.service.conceptcollection.mapper.IConceptCollectionDeepMapper;
 import edu.asu.spring.quadriga.service.workbench.mapper.IProjectShallowMapper;
 import edu.asu.spring.quadriga.service.workspace.IListWSManager;
+import mockit.Mocked;
 
 public class ConceptCollectionManagerTest {
 
-	@Mock
-	private RestTemplate mockedRestTemplate = Mockito.mock(RestTemplate.class);
+    @Mock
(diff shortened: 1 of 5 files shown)


jdamerow:
what final classes did you have to mock?

jdamerow:
ok to test

suprajkolluri:
String.class

jdamerow:
where do you mock that?

jdamerow:
ok, never mind, I saw it.

Quadriga/src/test/java/edu/asu/spring/quadriga/service/impl/conceptcollection/ConceptCollectionManagerTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+        List<ConceptpowerReply.ConceptEntry> conceptEntries = new ArrayList<ConceptpowerReply.ConceptEntry>();
+        ConceptpowerReply.ConceptEntry entry = new ConceptpowerReply.ConceptEntry();
+        entry.setLemma("test-lemma");
+        conceptEntries.add(entry);
+        ConceptpowerReply rep = new ConceptpowerReply();
+        rep.setConceptEntry(conceptEntries);
+
+        Mockito.when(mockedRestTemplate.getForObject(Matchers.eq(conceptURL + searchURL + "{name}/{pos}"),
+                Matchers.eq(ConceptpowerReply.class), Matchers.anyMapOf(String.class, String.class))).thenReturn(rep);
```
jdamerow: actually, the third argument is where you need to provide id/name/pos so that the rest template returns the correct answer for the given name/pos or id. sorry, I didn't look closely enough before.

suprajkolluri: Made the change and replaced all Mockito.anyString() with exact string matchers.

suprajkolluri:
Actually Jmockit was not necessary, mocked the string values using spring ReflectionUtils. Removed the corresponding dependency.

jdamerow:
ok to test

PR #58, opened 2015-12-16
Story/quad 44

suprajkolluri:
Added basic auth for rest calls to qstore from quadriga

### DIFF
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkManager.java
@@ -5,6 +5,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringWriter;
 import java.nio.ByteBuffer;
+import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkManager.java
```diff
@@ -786,6 +796,13 @@ public String storeXMLQStore(String XML) throws ParserConfigurationException, SA
 		}
 		return res;
 	}
+	
+	private String getAuthHeader(){
+	    UserDetails user = (UserDetails) SecurityContextHolder.getContext().getAuthentication().getPrincipal();
+        String auth = user.getUsername() + ":" + user.getPassword();
+        byte[] encodedAuth = Base64.encodeBase64( auth.getBytes(Charset.forName("US-ASCII")) );
+        return "Basic " + new String( encodedAuth );
```
jdamerow, line 804: the user for qstore should be configured in a properties file with variables filled through pom.xml

jdamerow, line 804: you can talk to karthik, he did something similar (you wouldn't need to use spring-security but the properties fiel with pom.xml is the same)

jdamerow:
see comment

PR #59, opened 2015-12-18
QUAD-50: Fixing the bug where dictionary items were not getting

suprajkolluri:
Fixing the bug where dictionary items were not getting retrieved. Replaced dictionaryDTO.getOwner() with dictionaryDTO.getOwner().getCreatedby()


jdamerow:
I think that's logically wrong. The owner can change, but the createdby should always stay the same if I'm not mistaken. Depending on what the logic is here, you need to change that.

PR #60, opened 2015-12-18
Bug/quad 50

suprajkolluri:
We are trying to show the dictionary details only to its owner. Modified the if condition. Please let me know if you think this is still incorrect.


jdamerow:
ok to test

jdamerow:
add to whitelist

PR #62, opened 2015-12-23
QUAD-41:  Made sure the submit button will be disabled until the user actually changes the value of the custom URL field.

suprajkolluri:
Made sure the submit button will be disabled until the user actually changes the value of the custom URL field.


jdamerow:
retest this please

PR #70, opened 2016-01-29
QUAD-54: 

suprajkolluri:
Fixed the null pointer exception by changing the model map attribute name from projectid to myprojectid as maven was overwriting
projectid's value.


jdamerow:
ok to test

PR #74, opened 2016-02-01
QUAD-53 : Adding back button to workbench.

kavinya:
QUAD-53 : Requesting pull request for - Addition of back button to Workspace Details page


location
```diff
@@ -0,0 +1 @@
+Subproject commit bafea4c19cbf87d821b9890430d45723ea95d84f
```
jdamerow, line 1: It looks like you added a submodule here. Please remove that before this can be merged into develop.

PR #75, opened 2016-02-01
QUAD-59 : Fixed bug on delete project button in workspace page.

kavinya:
Requesting pull request for QUAD-59 : Fixed bug on delete project buttons in workspace page


location
```diff
@@ -0,0 +1 @@
+Subproject commit bafea4c19cbf87d821b9890430d45723ea95d84f
```
jdamerow, line 1: Same here. Why is there a submodule?

PR #79, opened 2016-02-03
Dspace Code Removed

charan1312:
@jdamerow  Worked on Story/QUAD-61. The changes are done. Pushed the code to sprint1/QUAD-61 created from story/QUAD-61 .

Creating pull request. Please let me if there are any issues.


jdamerow:
ok to test

PR #82, opened 2016-02-03
Story/quad 63

ajaymodi:
sprint 1 updates


jdamerow:
ok. for the coming sprint please create a branch story/QUAD-63/sprint2.

PR #86, opened 2016-02-05
Story/quad 58

rajataggarwal91:
There was some code in workspacedetails.jsp related to dspace, which was obsolete. This change removes that code. This is the second change related to "Workspace Details Page: remove DSpace buttons"


jdamerow:
ok to test

jdamerow:
Looks like you broke the page. I get the following error:
org.apache.jasper.JasperException: /WEB-INF/views/workspace/workspacedetails.jsp(92,8) The end tag "&lt;/c:when" is unbalanced

PR #95, opened 2016-02-09
Story/quad 58

rajataggarwal91:
Another pull request for QUAD-58 against the previous one which was rejected. It is strange that I did not get any compilation or run time exceptions on my machine while running the project. But I see where the extra erroneous tag was present and that has been removed.

On top of that wherever <choose> and <when> were present related to dSpaceKeys, they were removed. So in these cases, the proceeding scripts have become unconditional. For example, 
<c:choose> <c:when test="${empty dspaceKeys}">
In the above line <c:choose> and <c:when> were removed and <script> after the tags was made unconditional.

The code has been revisited twice to check for any missing tags. If there is still some code breakage, please let me know and I will look into the issue once more. I think it might be due to no JSP validation or different environment on your machine that causes the exceptions to show up on your machine but not mine.


jdamerow:
ok to test

PR #98, opened 2016-02-10
Story/quad 63

ajaymodi:
We have changed code according to suggestion with proper formatting. Please review and merge.


jdamerow:
please resolve conflicts

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/ConceptCollectionRestController.java
```diff
@@ -448,5 +465,60 @@
...
+                        throw new QuadrigaException(e.getMessage(), e);
+                    }
+
+                }
+            }
+            
+            return new ResponseEntity<String>(ja.toString(), HttpStatus.OK);
+            
+        } catch (QuadrigaStorageException e) {
+            throw new QuadrigaStorageException();
```
jdamerow, line 521: the try/catch block should be where the exception is actually thrown. If one collection can't be retrieved but the others can, you want to show the the once that you could retrieve and just show an error. Also, given that this is a method that returns JSON, return proper error codes and useful error messages. And don't swallow the exception ;)

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/DictionaryRestController.java
```diff
@@ -193,7 +212,7 @@
      * @param dictionaryId
      * @param model
      * @return
-     * @throws RestException
+     * @throws Exception
```
jdamerow, line 215: why did you change that?

jdamerow, line 215: and the ones above?

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/DictionaryRestController.java
```diff
@@ -404,5 +423,61 @@
...
+            QuadrigaException {
+        String userId = principal.getName();
+        List<IProjectDictionary> dictionaryList = null;
+        try {
+            // TODO: listProjectDictionary() is to be changed according to
+            // mapper
+            dictionaryList = projectDictionaryManager.listProjectDictionary(
+                    projectid, userId);
+            
+            dictionaryList.removeAll(Collections.singleton(null));
```
jdamerow, line 457: why are you doing this? there shouldn't be any null in the list, or did you have that case?

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/DictionaryRestController.java
```diff
@@ -404,5 +423,61 @@
...
+                        throw new QuadrigaException(e.getMessage(), e);
+                    }
+                }
+            }
+                 
+            return new ResponseEntity<String>(ja.toString(), HttpStatus.OK);
+
+        } catch (QuadrigaStorageException e) {
+            throw new QuadrigaStorageException();
+        }
```
jdamerow, line 478: same as before, try/catch needs to be where it's actually thrown. and proper error handling is needed.

ajaymodi, line 478: So when you say, proper error handling should we just output status code like 200, 404 or you still want it to have quadrigaexception?

jdamerow, line 478: yes, return have it return something that can be used by the client, with the correct status code and  an error message. In other words, catch the exception and return a ResponseEntity.

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/DictionaryRestController.java
ajaymodi, line 457: This code was already there in dictionarycontroller, we are just creating Json Object and returning it.

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/DictionaryRestController.java
ajaymodi, line 215: This was a mistake from our side due to lot of merge conflicts. Let me know if we still miss anything.

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/DictionaryRestController.java
jdamerow, line 457: just remove it unless you get errors when you remove it, but I don't think we need this. If it's necessary the underlying code should be changed to make sure the list doesn't have any nulls in it.

jdamerow:
please resolve conflicts

ajaymodi:
please check pull request from story quad-62. We have merged 63 into 62.

PR #105, opened 2016-02-15
Update navbar.jsp

sarvani11:
Story QUAD- 88


jdamerow:
ok to test

jdamerow:
This shouldn't look like this:
image
The navigation bar should keep the same height.

sarvani11:
Will make the changes

PR #106, opened 2016-02-15
QUAD-98 : Added sites page to quadriga

charan1312:


jdamerow:
ok to test

Quadriga/src/main/java/edu/asu/spring/quadriga/web/HomeController.java
```diff
@@ -122,6 +122,24 @@ public String home(Locale locale, Model model, Principal principal)
...
     }
 
+    /**
+     * This method is used to access the public page -sites that enlists the public pages of all the projects 
+     * @param locale
+     * @param model
+     * @return
+     */
+    @RequestMapping(value = "sites", method = RequestMethod.GET)
+    public String showQuadrigaPublicPages(Locale locale, Model model) {
```
jdamerow, line 132: This method should be in its own controller.

jdamerow:
In the menu bar, remove the Networks, Browse, and Search button for now and add "Quadriga Network Projects" on the left side of the navigation:
image

PR #109, opened 2016-02-16
Story/quad 80

jdamerow:


jdamerow:
@charan1312 please move the corresponding jira ticket to done.

PR #110, opened 2016-02-17
Story/quad 65

JNagarkar:
Changed the layout of skeleton_public page and added attributes in tiles-def.xml for displaying networks in public pages.


Quadriga/pom.xml
```diff
@@ -526,10 +526,10 @@
...
-		<connection>scm:svn:http://none</connection>
-		<url>scm:svn:https://none</url>
-	</scm>
+<!-- 	<scm> -->
+<!-- 		<connection>scm:svn:http://none</connection> -->
+<!-- 		<url>scm:svn:https://none</url> -->
+<!-- 	</scm> -->
 
 	<reporting>
 		<plugins>
```
jdamerow, line 537: please don't submit your modified pom.xml

Quadriga/src/main/webapp/WEB-INF/tiles/layouts/skeleton_public.jsp
```diff
@@ -48,17 +48,51 @@
...
+								<!-- Sidebar -->
+								<!-- Recent Posts -->
+								<tiles:insertAttribute name="right-navigation" />
+								<!-- /Recent Posts -->
+								<!-- /Sidebar -->
+							</div>
+						</div> 
+					</div> <!-- end of row -->
+ 
+    </div>
```
jdamerow: You don't want to change the layout for all public pages. Most other pages don't have a side navigation and you don't want to use that space for nothing. Just create a new skeleton for public pages with side navigation and use that for the network page instead of the general one.

JNagarkar:
reverted the public page to its original version, created a new skeleton page for networks as skeleton_public_network.jsp .

JNagarkar:
Made changes in tiles-def.xml and changed tag classes in skeleton_public_network.jsp to bootstrap classes, but publicvisualnetwork.jsp uses predefined classes for displaying, so page created is not completely responsive.

jdamerow:
ok to test

jdamerow:
the page should use the same font as the other pages and the same padding:
image
compare to:
image

jdamerow:
also the backgrounds should be white

jvutukur:
There is a slight change in layout because of loading unnecessary css files. Removed those unnecessary css file imports. Fixed the issues..

jdamerow:
retest this please

PR #112, opened 2016-02-17
QUAD77- network fullscreen option

rmanojcit:
QUAD77- network fullscreen option


jdamerow:
ok to test

jdamerow:
three things:
- make the button an icon (e.g. <i class="fa fa-arrows-alt"></i>) and put it in the upper left or right corner
- the div should take up the whole screen width (otherwise the full screen doesn't make much sense)
- the icon should also be on the public pages for the networks

jdamerow:
please reopen this pull request when you made the changes?

PR #118, opened 2016-02-17
QUAD-93: Fixed the bug

suprajkolluri:


jdamerow:
@suprajkolluri, I removed the whole javascript approval process and replace it with a backing bean and a post method. This will give us less trouble and is easier to main in the future. What's missing right now is input validation. I'll try to finish it tomorrow, but if I don't get around to it would you mind finishing it? Thanks, Julia

jdamerow:
ok to test

PR #125, opened 2016-02-29
Story/QUAD-97 Story/QUAD-99

v1kd:
Story 97 - To search QStore4S to for the concept
Story 99 - Show the resulting network

### DIFF
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
@@ -9,6 +9,7 @@ import edu.asu.spring.quadriga.transform.Node;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.security.crypto.keygen.StringKeyGenerator;
 import org.springframework.stereotype.Service;
 
 import edu.asu.spring.quadriga.domain.network.INetworkNodeInfo;
@@ -111,7 +112,7 @@ public class NetworkTransformationManager implements INetworkTransformationManag
                     // search for concept id
                     Node searchNode = null;
                     for (Node node: transformedNetwork.getNodes().values()) {
-                        if (node.getConceptId().contains(conceptId)) {
+                        if (node.getConceptId().equals(conceptId)) {
                             searchNode = node;
                             break;
                         }
@@ -121,19 +122,24 @@ public class NetworkTransformationManager implements INetworkTransformationManag
                         continue;
                     }
 
-                    String statementId = searchNode.getStatementIds().get(0);
+                    Set<String> searchStatements = new HashSet<String>(searchNode.getStatementIds());
+                    // String statementId = searchNode.getStatementIds().get(0);
                     Set<Node> associatedNodes = new HashSet<Node>();
                     for (Map.Entry<String, Node> entry: transformedNetwork.getNodes().entrySet()) {
                         // check if first statementId is in all the nodes
                         Node n = entry.getValue();
-                        if (n.getStatementIds().contains(statementId)) {
-                            searchNodes.put(entry.getKey(), n);
-                            associatedNodes.add(n);
+                        for (String statementId: n.getStatementIds()) {
+                            if (searchStatements.contains(statementId)) {
+                                searchNodes.put(entry.getKey(), n);
+                                associatedNodes.add(n);
+                            }
                         }
                     }
 
                     // find all the links associated with the searched nodes
+                    // using depth first search
                     for (Link link: transformedNetwork.getLinks()) {
+                        // it may be in source or target
                         if (associatedNodes.contains(link.getSubject())) {
                             searchLinks.add(link);
                         }
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
@@ -97,7 +97,7 @@ public class NetworkTransformationManager implements INetworkTransformationManag
         if (networkList == null) {
             return null;
         }
-        
+
         Map<String, Node> searchNodes = new HashMap<String, Node>();
         List<Link> searchLinks = new ArrayList<Link>();
         for (INetwork network : networkList) {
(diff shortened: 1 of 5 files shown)


Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +19,145 @@
...
+                List<INetworkNodeInfo> localNetworkNodeInfoList =
+                        networkManager.getNetworkTopNodes(network.getNetworkId());
+                // check if this network has the node with given concept id
+                if (localNetworkNodeInfoList != null) {
+                    ITransformedNetwork transformedNetwork =
+                            transformer.transformNetwork(localNetworkNodeInfoList);
+                    // search for concept id
+                    Node searchNode = null;
+                    for (Node node: transformedNetwork.getNodes().values()) {
+                        if (node.getConceptId().contains(conceptId)) {
```
jdamerow: that should be equal I believe

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +19,145 @@
...
+                            searchNode = node;
+                            break;
+                        }
+                    }
+                    // if node does not exist continue
+                    if (searchNode == null) {
+                        continue;
+                    }
+
+                    String statementId = searchNode.getStatementIds().get(0);
```
jdamerow: why only the first one?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +19,145 @@
...
+                    }
+
+                    // find all the links associated with the searched nodes
+                    for (Link link: transformedNetwork.getLinks()) {
+                        if (associatedNodes.contains(link.getSubject())) {
+                            searchLinks.add(link);
+                        }
+                    }
+                }
+            } catch (QuadrigaStorageException e) {
```
jdamerow, line 148: put this where it is thrown

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +19,145 @@
...
+                        network.getNetworkId(), e);
+            }
+        }
+
+        ITransformedNetwork networkJSon = transformer.transformNetwork(networkNodeInfoList);
+        return networkJSon;
+    }
+
+    @Override
+    public ITransformedNetwork searchTransformedNetworkOfProject(String projectId, String conceptId)
```
jdamerow, line 95: and write a test class for this :)

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +20,150 @@
...
+                // check if this network has the node with given concept id
+                if (localNetworkNodeInfoList != null) {
+                    ITransformedNetwork transformedNetwork =
+                            transformer.transformNetwork(localNetworkNodeInfoList);
+                    // search for concept id
+                    Node searchNode = null;
+                    for (Node node: transformedNetwork.getNodes().values()) {
+                        if (node.getConceptId().equals(conceptId)) {
+                            searchNode = node;
+                            break;
```
jdamerow, line 117: hm, so wouldn't that get only one node with the given concept form each network? But if I search for a predicate (e.g. engages with), I might have several nodes with that concept. Or am I missing something?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +20,150 @@
...
+                    for (Map.Entry<String, Node> entry: transformedNetwork.getNodes().entrySet()) {
+                        // check if first statementId is in all the nodes
+                        Node n = entry.getValue();
+                        for (String statementId: n.getStatementIds()) {
+                            if (searchStatements.contains(statementId)) {
+                                searchNodes.put(entry.getKey(), n);
+                                associatedNodes.add(n);
+                            }
+                        }
+                    }
```
jdamerow, line 137: Hm, so here is what I would do to avoid having to do this for every node in every network:
1. modify the class TransformedNetwork to also have a map that holds the nodes for each statementid in the network (this might be useful for other things anyway).
2. remove the setNodes method for TransformedNetwork and have instead a addNode method that adds a node to a list and also adds the node to the list of the corresponding statement id
3. add a "clear" method to remove all nodes
4. make the getNodes method return an immutable list 
5. adjust the places where nodes are added to the network to use the new methods
6. use the map<statementId, nodeslist> here instead of these two for loops

this will make your code shorter and easier to read

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +20,150 @@
...
+
+                    // find all the links associated with the searched nodes
+                    // using depth first search
+                    for (Link link: transformedNetwork.getLinks()) {
+                        // it may be in source or target
+                        if (associatedNodes.contains(link.getSubject())) {
+                            searchLinks.add(link);
```
jdamerow, line 144: we talked about this

jdamerow:
please reopen this PR when you made the changes

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
v1kd, line 137: The problem is TransformedNetwork class has not been setup properly. Even if I had a map with statement id and list of nodes, I still have to search in all the entries in the map for a given concept id.

A lot of code has to be changed in order to fix the TransformedNetwork class. I have come up with a different solution which is a lot similar to yours.
1. Get the combined network of the project
2. Add all the nodes which has a given concept id in a list
3. I will reduce the nodes to map<string, set<Node>> where string is the statement id and node is the one which is included in that statement.
4. Now with this I can get all the statements which contains the concept id
5. I will remove other nodes and links from the transformed network.
6. Please refer to this pull request - https://github.com/diging/quadriga/pull/143

PR #130, opened 2016-03-01
Story/quad 114

ajaymodi:


ajaymodi:
quad-114

Any idea how to fix it?

ajaymodi:
Getting attribute local is not allowed to appear in ref error while merging develop branch.

jdamerow:
Do you have all the namespaces and schemas included at the top of the file?

ajaymodi:
Yes, I face this issue when I merge develop branch with story branch. I even checked out develop branch to see whether is it working? I faced the same problem with develop branch.

http://stackoverflow.com/questions/21054171/spring-beans-xsd-does-not-define-local-why  
According to this link, I changed local to bean and now css is not loading up.

image

jdamerow:
the above screen shot look like a maven issue. Have you set the Maven profile?

ajaymodi:
@Override
    public String getConceptLemmaFromConceptId(String id) {

```
    Map<String, String> vars = new HashMap<String, String>();
    vars.put("name", id.trim());
    String lemma = id;
    ConceptpowerReply rep = conceptpowerConnector.getById(id);
    if (rep.getConceptEntry().size() == 0) {
        return lemma;
    }
    return rep.getConceptEntry().get(0).getLemma();
}
```

This method is in conceptcollectionmanager. Why are we using vars hashmap here? It seems to me a local variable which is not used further.

ajaymodi:
I have modified code to as per your suggestion. Please retest it.

jdamerow:
Jup, the hashMap is a remnant from before refactoring. You can just delete it.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/conceptcollection/IConceptCollectionManager.java
```diff
@@ -127,5 +127,8 @@ void addItems(String lemmma, String id, String pos, String desc,
 
 
 	public abstract String getProjectsTree(String userName,String ccId) throws JSONException;
+
+
+    String getConceptDescriptionFromConceptId(String id);
```
jdamerow, line 132: Please format this class so that all method declaration start at the same position. Thanks.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/impl/conceptcollection/ConceptCollectionManager.java
```diff
@@ -202,6 +202,27 @@ public String getConceptLemmaFromConceptId(String id) {
...
+     * @param id
+     *            - item id
+     * @return String - Description associated with concept
+     */
+    @Override
+    public String getConceptDescriptionFromConceptId(String id) {
+
+        Map<String, String> vars = new HashMap<String, String>();
+        vars.put("name", id.trim());
+        String desc = id;
```
jdamerow, line 216: Jup, just delete the map.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/WebsiteProjectController.java
```diff
@@ -215,5 +224,104 @@ public String visualizeAllNetworks(@PathVariable("projectUnixName") String proje
 
 		return "sites/networks/visualize";
 	}
+	
+	public static JSONArray sortByValue(Map<String, Integer> map, Map<String, String[]> map2) throws JSONException {
```
jdamerow, line 228: A few things:
1. formatting
2. why is this a static method?
3. I think you can delete this methods. See later comments.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/WebsiteProjectController.java
```diff
@@ -215,5 +224,104 @@ public String visualizeAllNetworks(@PathVariable("projectUnixName") String proje
...
+	 * This method gives the visualization of  how often concepts appear in the networks
+	 * @author Bharath Srikantan & Ajay Modi
+	 * @param projectUnixName	The project unix name
+	 * @param model				Model
+	 * @return view
+	 * @throws JAXBException
+	 * @throws QuadrigaStorageException
+	 */
+	@RequestMapping(value = "sites/{projectUnixName}/statistics", method = RequestMethod.GET)
+	public String showProjectStatistics(@PathVariable("projectUnixName") String projectUnixName,
```
jdamerow, line 263: formatting

jdamerow, line 263: this method should be in its own controller

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/WebsiteProjectController.java
```diff
@@ -215,5 +224,104 @@ public String visualizeAllNetworks(@PathVariable("projectUnixName") String proje
...
+                        String lbl = s.getLabel(); 
+                        if(top10Concepts.containsKey(url)){
+                            top10Concepts.put(url, top10Concepts.get(url)+1);
+                        }else{
+                            top10Concepts.put(url, 1);
+                            top10ConceptsDetails.put(url, new String[] {des, lbl});
+                        }
+                    }
+                    
+                }
```
jdamerow, line 301: This is all business logic and should not be in the controller. You should create a new service class to do your calculations and then your controller should only worry about getting the data to be represented and representation stuff.

jdamerow, line 301: Also, create a class to hold the information you need for the statistics (like a concept stats class holding information about a concept and how often it appears), you can then return an already sorted list. Your controller then only needs to worry about creating JSON from the list.

jdamerow, line 301: And the number of how many concepts should be shown in the statistics should not be hardcoded but configurable (for now through a system property, later we probably want to make that configurable by project).

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/WebsiteProjectController.java
```diff
@@ -215,5 +224,104 @@ public String visualizeAllNetworks(@PathVariable("projectUnixName") String proje
...
+             
+            }
+            try {
+                jo = sortByValue(top10Concepts, top10ConceptsDetails);
+                model.addAttribute("show_success_alert", true);
+                successMsg.append("Top10 Concepts from network fetched successfully");
+                successMsg.append("\n");
+                System.out.print(jo.toString());
```
jdamerow, line 309: Don't show any success message, it's just a stats display page, no interaction from the user that would need a success message. Also, remove the sysout.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/WebsiteProjectController.java
```diff
@@ -215,5 +224,104 @@ public String visualizeAllNetworks(@PathVariable("projectUnixName") String proje
...
+                errorMsg.append("\n");
+                
+            }
+        }
+
+        model.addAttribute("success_alert_msg", successMsg.toString());
+        model.addAttribute("error_alert_msg", errorMsg.toString());
+        
+		return "sites/project/statistics";
+	}
```
jdamerow, line 326: what's up with all those indentation issues?

jdamerow:
please reopen PR when you've made the changes.

ajaymodi:
Will you please tell me what indentation style are we using? Is it tab 4 or space 4 or none of it?

jdamerow:
https://diging.atlassian.net/wiki/pages/viewpage.action?pageId=8781898
4 spaces

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/WebsiteProjectController.java
ajaymodi, line 326: I was using tab:4 indentation, now I have changed it to spaces:4. Sometimes, formatting with eclipse does not match up with how git shows, however I am not sure which one shows the correct one in that case.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/WebsiteProjectController.java
ajaymodi, line 228: Is it a bad practice to create a static method?

never mind, got the mistake.

jdamerow, line 228: yes, it is also bad practice and unnecessary.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/WebsiteProjectController.java
jdamerow, line 326: I think it's when you mix tabs and spaces when Github gets confused. Yes, 4 spaces should do it.

PR #134, opened 2016-03-03
Quad 105

jvutukur:
Changed the size of pop ups and its background color. Used class form-control of bootstrap

capture


jdamerow:
ok to test

PR #140, opened 2016-03-10
Story/quad 74

v1kd:
To show combined network in a project.
Now there would be only one subject and object nodes in the entire network.

### DIFF
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
@@ -23,75 +23,75 @@ import edu.asu.spring.quadriga.service.network.domain.ITransformedNetwork;
 @Service
 public class NetworkTransformationManager implements INetworkTransformationManager {
 
-	private static final Logger logger = LoggerFactory.getLogger(NetworkTransformationManager.class);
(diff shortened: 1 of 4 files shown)


Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +23,127 @@
...
+
+        Map<String, Node> updatedNodes = new HashMap<String, Node>();
+        int index = 0;
+        for (Node node: nodes.values()) {
+            // if the node is predicate node
+            // add to the updated nodes map
+            if (node instanceof PredicateNode) {
+                // the key does not matter
+                // as it is never being used
+                updatedNodes.put(String.valueOf(++index), node);
```
jdamerow: Is there a reason you create a new key here instead of using the one that already exists (the one in the nodes map)?

v1kd: As the keys are never used to compute any value I decided to use a different keys. I changed to the keys to the keys which are in the original map. Please review and merge.

v1kd:
Changed the keys to the updated node map to the original keys. Please review.

jdamerow:
ok to test

PR #143, opened 2016-03-11
Story/quad 99 Story/quad 97

v1kd:
New approach to search for the node given a concept id and return the network.

### DIFF
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
@@ -154,27 +154,34 @@ public class NetworkTransformationManager implements INetworkTransformationManag
             }
         }
 
-        // Include only those nodes which have their statement ids
-        // in statement search set
-        Map<String, Node> finalNodes = new HashMap<String, Node>();
-        for (Map.Entry<String, Node> entry: transformedNetwork.getNodes().entrySet()) {
-            Node node = entry.getValue();
-            for (String statementId: node.getStatementIds()) {
-                if (statementIdSearchSet.contains(statementId)) {
-                    // statement id match. Add it to the final nodes map
-                    finalNodes.put(entry.getKey(), node);
-                }
-            }
-        }
-
-        // Include only those links which have their statement ids
-        // in statement search set
+        // include only those links which have statement ids in the search set
         List<Link> finalLinks = new ArrayList<Link>();
         for (Link link: transformedNetwork.getLinks()) {
             if (statementIdSearchSet.contains(link.getStatementId())) {
                 // statement id match
                 // add to the final link list
                 finalLinks.add(link);
+            }
+        }
+
+        // final nodes
+        Map<String, Node> finalNodes = new HashMap<String, Node>();
+        Set<Node> addedNodes = new HashSet<Node>();
+        Integer index = 0;
+        for (Link link: finalLinks) {
+            Node subjectNode = link.getSubject();
+            Node objectNode = link.getObject();
+            // if node already added then do not add it
+            // if nodes are added twice - it would produce many
+            // nodes and less links => disconnected graph
+            if (!addedNodes.contains(subjectNode)) {
+                finalNodes.put((++index).toString(), subjectNode);
+                addedNodes.add(subjectNode);
+            }
+
+            if (!addedNodes.contains(objectNode)) {
+                finalNodes.put((++index).toString(), objectNode);
+                addedNodes.add(objectNode);
             }
         }
 
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
@@ -10,7 +10,6 @@ import edu.asu.spring.quadriga.transform.PredicateNode;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.security.access.method.P;
 import org.springframework.stereotype.Service;
 
 import edu.asu.spring.quadriga.domain.network.INetworkNodeInfo;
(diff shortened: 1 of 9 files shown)


Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +21,205 @@
...
+                it.remove();
+            } else {
+                searchedNodeSet.addAll(nodeSet);
+            }
+        }
+
+        // now include the statements
+        Map<String, Node> finalNodes = new HashMap<String, Node>();
+        int index = 0;
+        for (Node n: searchedNodeSet) {
```
jdamerow: bad variable name

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +21,205 @@
...
+                    isNodeExists = true;
+                }
+            }
+            // remove the statement if it is not present
+            if (!isNodeExists) {
+                it.remove();
+            } else {
+                searchedNodeSet.addAll(nodeSet);
+            }
+        }
```
jdamerow, line 189: I think you can simplify this a lot by using the statement ids:
1. instead of have a list of search nodes, have a list of statement ids (every node should have a list of statement ids it belongs to).
2. then you can just take all the sets from your statementIdMap for all statement ids that you just found and put them in one big set. that should make this whole second step much more compact.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +21,205 @@
...
+        Iterator<Link> linkIterator = finalLinks.iterator();
+        while (linkIterator.hasNext()) {
+            Link link = linkIterator.next();
+            // if the subject or object does not exist in any links
+            // remove that link
+            if (!searchedNodeSet.contains(link.getSubject()) ||
+                    !searchedNodeSet.contains(link.getObject())) {
+                linkIterator.remove();
+            }
+        }
```
jdamerow, line 196: If you add a statement id field to the Link class and modify the eventparser accordingly then you can do the same thing (as described above, filtering by statement id) with links.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
```diff
@@ -96,4 +106,30 @@ public String getSearch(@PathVariable("projectUnixName") String projectUnixName,
         return new ResponseEntity<String>(defaultJsonErrorMsg, HttpStatus.INTERNAL_SERVER_ERROR);
     }
 
+    @RequestMapping(value = "sites/{projectUnixName}/networks/search", method = RequestMethod.GET)
+    public String getSearchTransformedNetwork(@PathVariable("projectUnixName") String projectUnixName,
+                                              @RequestParam("conceptId") String conceptId,
+                                              Model model)
+        throws JAXBException, QuadrigaStorageException {
```
jdamerow: where is the JAXBException thrown? You should handle it and not just pass it on.

jdamerow:
see comments. please reopen this PR when you're done.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
v1kd: It does not throw anywhere. I removed it.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
v1kd, line 189: It makes sense. This is the final change I made.
1. I search for all the nodes which I whose concept id matches the search concept id and then store all the statement ids in a hash set.
2. I will iterate over all the nodes and check if there is a statement that matches the search statement id set.
3. If it matches I will add that to a final node map.
4. I added a statement id field in Link class.
5. I will iterate through all the links and check if the statement id exists in the statement id set.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
v1kd: Changed

v1kd:
Updated the code as per the comments.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +20,191 @@
...
+        List<INetworkNodeInfo> oldNetworkTopNodesList = null;
+
+        try {
+            oldNetworkTopNodesList = networkManager.getNetworkTopNodesByVersion(networkId, Integer.parseInt(versionID));
+        } catch (QuadrigaStorageException e) {
+            logger.error("DB Error while getting network top nodes", e);
+        }
+
+        networkJSon = transformer.transformNetwork(oldNetworkTopNodesList);
+        return networkJSon;
```
jdamerow, line 47: merge with line 44 and remove line 35.

jdamerow, line 47: never mind. I don't think it's your code.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +20,191 @@
...
+            // nodes and less links => disconnected graph
+            if (!addedNodes.contains(subjectNode)) {
+                finalNodes.put((++index).toString(), subjectNode);
+                addedNodes.add(subjectNode);
+            }
+
+            if (!addedNodes.contains(objectNode)) {
+                finalNodes.put((++index).toString(), objectNode);
+                addedNodes.add(objectNode);
+            }
```
jdamerow, line 188: I think you can move up what you're doing here in the loop over finalLinks to your last loop (line 161-164).

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -17,79 +20,191 @@
...
+        }
+
+        // finally add the searched concept nodes if they are not present
+        for (Node node: searchedNodes) {
+            if (!searchedNodes.contains(node)) {
+                finalNodes.put((++index).toString(), node);
```
jdamerow: why are you creating a new Id?

v1kd: If I do not use another unique id then I have to search the entire map for the key and then add the node with key in the final map. In fact we actually do not use this key anywhere in the program. When you look into `D3Creator.java` class (which would take the nodes and links and create a json out of it) the first statement is 
`List<Node> nodeList = new ArrayList<Node>(nodes.values());`
And the entire code would just use the same `nodeList`

And couple of things regarding `D3Creator.java`
1. It does not use a JSON creator so the json message it constructs would be invalid when the node attributes contain a double quote string.
2. You can see it uses `nodeList.indexOf(link.getObject())` code in couple of places which would search for the entire array list each and every time which is a O(n*n) solution

A better approach would be take the links and iterator over the links. For each link add the subject and object to a list node and we already know the indexes of these two nodes. The time complexity would be O(n).

Please let me know if you think refactoring is necessary. I can work on it.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
v1kd, line 188: I first did that way but a lot was happening in the inside for loop. I thought it was easy to read the code this way. I changed the code as per your suggestion.

jdamerow:
ok to test

jdamerow:
The probably 2 last things:
- when I hit enter after typing something in (before selecting an entry from the dropdown), I get a Tomcat error page.
- when there are no results for my search, the page should tell me that there are no results. Right now it's just a blank network page.

v1kd:
Fixed the tomcat error page.
Updated the results page to show information if the network results are empty. Please review and merge.

v1kd:
I added the search term label and description. Please review.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
```diff
@@ -96,4 +110,51 @@ public String getSearch(@PathVariable("projectUnixName") String projectUnixName,
...
+        }
+
+        String searchNodeLabel = "";
+        if (transformedNetwork == null || transformedNetwork.getNodes().size() == 0) {
+            model.addAttribute("isNetworkEmpty", true);
+        } else {
+            // get the search concept id
+            for (Node node: transformedNetwork.getNodes().values()) {
+                if (conceptId.equals(node.getConceptId())) {
+                    searchNodeLabel = node.getLabel();
```
jdamerow: Why don't you get the label from the conceptpower reply (when you get the description)? It's unnecessary to iterate over all the nodes again.

Quadriga/src/main/webapp/WEB-INF/views/navigation/nav-networkrightsearchplane.jsp
```diff
@@ -0,0 +1,100 @@
...
+		margin: 0px;
+		opacity: 0.7;
+	}
+		
+	table,th,td{
+		border:1px solid black;
+	}
+
+		
+	}
```
jdamerow: ?

jdamerow:
Great, just these few small things. Good work!

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
v1kd: I removed the for loop and took the label for conceptpowerReply object.

Quadriga/src/main/webapp/WEB-INF/views/navigation/nav-networkrightsearchplane.jsp
v1kd: I have removed unnecessary styles. The page is only using 'textarea' style

PR #148, opened 2016-03-16
Story/quad 100 sprint3

charan1312:
@jdamerow : QUAD-100-sprint3: Creating pull request into QUAD-98 branch from which it is forked.


jdamerow:
Please modify/commit only the files you actually changed. This PR has 127 files. I don't believe you modified all of them. Did you do a general reformatting? The way it is right now, I can't see what code you actually changed.

PR #158, opened 2016-03-18
QUAD-134: Replacing the resource file name to projectconstants

suprajkolluri:


Quadriga/src/main/java/edu/asu/spring/quadriga/dao/impl/NetworkDAO.java
```diff
@@ -62,7 +62,7 @@
     @Autowired
     private NetworkDTOMapper networkMapper;
 
-    @Resource(name = "database_error_msgs")
+    @Resource(name = "projectconstants")
     private Properties messages;
```
jdamerow: ah that makes sense. Would you please just delete this? It's in the parent class.

suprajkolluri:
Just did it

Quadriga/src/main/java/edu/asu/spring/quadriga/dao/impl/NetworkDAO.java
suprajkolluri: Just did it.

jdamerow:
ok to test

PR #169, opened 2016-03-23
Bug/quad 129

suprajkolluri:


Quadriga/src/main/java/edu/asu/spring/quadriga/dao/impl/workspace/WorkspaceCCDAO.java
```diff
@@ -97,11 +96,8 @@ public String addWorkspaceCC(String workspaceId, String CCId, String userId) thr
 		}
 		catch(Exception e)
 		{
-			errMsg = "Oops! Error from DB";
-			logger.error("addWorkspaceCC method :",e);
         	throw new QuadrigaStorageException();
```
jdamerow: Please catch the specific exception being thrown and initialize the new exception with the one being thrown. otherwise we loose the stacktrace.

jdamerow:
ok to test

jdamerow:
There still seem to be several issues here:
1. when I go to add dictionary (no dictionaries have been added yet), click "add dictionary" without selecting one first, I see another page telling me no dictionary was selected. when I click ok, I see a blank page:
image
-> besides the empty page, ideally a flash message would be shown telling me to select a dictionary on the page where I can select a dictionary and click "add dictionary"
1. when I do the same thing with concept collections, I get redirected to the project page after clicking ok. Same thing here: ideally I see a flash message on the page where I select a concept collection that tells me I need to select one.
2. after adding a concept collection or dictionary, I should see a flash message telling me it was a success on the project details page. Right now I see this, which doesn't make much sense (it tells me it was successful but also that no CC was found):
   image

Please reopen this PR when you're done.

suprajkolluri:
Oh..this is for adding dictionaries and concepts from a Project. The bug was raised for adding them from a workspace. I will create a similar bug for a Project as well and fix it.

jdamerow:
ok to test

jdamerow:
ok, ya, that' makes sense. One last thing though, above point number two is true here as well. After adding a dictionary/concept collection the user should be redirected to the workspace page (and the flash message should be shown). Right now it does this:
image

Same thing for deleting concept collections/dictionaries. Afterwards it should redirect to the workspace page.

PR #176, opened 2016-03-28
Story/quad 126

v1kd:


Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/networks/CreationEvent.java
```diff
@@ -105,4 +105,19 @@
...
+    public String getSourceReference() {
+        for (JAXBElement element: getIdOrCreatorOrCreationDate()) {
+            if ("source_reference".equals(element.getName().getLocalPart())) {
+                return (String) element.getValue();
+            }
+        }
+
+        // send an empty string if it cannot find a source_reference
+        return "";
+    }
```
jdamerow, line 105: I know that's how it's done for other properties, but that's really not the best way and I'll probably refactor it eventually. You can simply add a String field "source_reference" to the class and remove the @XmlElementRef for the idOrCreatorOrCreationDate list. Just make sure that "source_reference" is not expected to be in the list by any method. Does this make sense?

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/networks/RelationType.java
```diff
@@ -133,44 +133,59 @@ public PredicateType getPredicateType(){
...
+    public String getSourceReference() {
+        for (JAXBElement element: idOrCreatorOrCreationDate) {
+            if ("source_reference".equals(element.getName().getLocalPart())) {
+                return (String) element.getValue();
+            }
+        }
+
+        // send an empty string if it cannot find a source_reference
+        return "";
+    }
```
jdamerow, line 174: same here

jdamerow:
see comment

v1kd:
Made changes as per suggestions. Please review.

jdamerow:
ok to test

PR #180, opened 2016-03-28
Story/quad 119

ajaymodi:
This is rough flow for 119 story. Please look into it and if any structural changes required, please let me know.

Thanks


Quadriga/src/main/java/edu/asu/spring/quadriga/service/publicwebsite/impl/ProjectStats.java
```diff
@@ -0,0 +1,144 @@
...
+                String username = network.getCreator().getUserName();
+                IUserStats userStats = mapUserNetworks.get(username);
+                if (userStats == null) {
+                    new UserStats(username, 0, 1);
+                } else {
+                    userStats.incrementNetworkCount();
+                }
+            }
+        }
+        return mapUserNetworks;
```
jdamerow, line 142: Since you pass mapUserNetworks into the method, you don't need to return it.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/ProjectStatsController.java
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        String projectId = project.getProjectId();
+        List<INetwork> networks = networkmanager
+                .getNetworksInProject(projectId);
+        List<IConceptStats> conceptsWithCount = null;
+        List<IUserStats> userStats; 
+        if (!networks.isEmpty()) {
+            conceptsWithCount = projectStats.getConceptCount(networks);
+            userStats = new ArrayList<IUserStats>(projectStats.getCountofNetwork(projectId, projectStats.getCountofWorkspace(projectId)).values());
```
jdamerow, line 105: Your projects stats service needs to be consistent. Right now there are three methods, all of them retrieving different statistics in different ways. One method gets a list of networks, the other two retrieve it themselves. One method returns a map, the next expects a map as input. Choose one strategy and follow it for the whole class. I would suggest, remove the dependency between your controller and the NetworkManager and let the project statistic service handle that part. Just give it projectid or username (whatever makes most sense) and let the stats service handle the rest. And then decide what you need: either you pass all your methods an object to fill, or if its more convenient have them each return a custom object or map or something with the results. does this make sense?

jdamerow, line 105: And definitely avoid these double calls to the same service using the result of one as input to the next! This is difficult to read and not very reusable.

Quadriga/src/main/webapp/WEB-INF/views/public/navbar.jsp
```diff
@@ -1,10 +1,6 @@
...
-        href="${pageContext.servletContext.contextPath}/sites/${project.unixName}/networks">Explore</a></li>
-<li ${currentPage == "networksearch" ? "class=\"active\"" : ""}><a
-        href="${pageContext.servletContext.contextPath}/sites/${project.unixName}/search">Search</a></li>
-<li ${currentPage == "getprojectstatistics" ? "class=\"active\"" : ""}><a
-        href="${pageContext.servletContext.contextPath}/sites/${project.unixName}/statistics">Statistics</a></li>
\ No newline at end of file
+<li ${currentPage == "home" ? "class=\"active\"" : ""}><a href="${pageContext.servletContext.contextPath}/sites/${project.unixName}">Home</a></li>
+<li ${currentPage == "networks" ? "class=\"active\"" : ""}><a href="${pageContext.servletContext.contextPath}/sites/${project.unixName}/browsenetworks">Networks</a></li>
+<li ${currentPage == "exploregraph" ? "class=\"active\"" : ""}><a href="${pageContext.servletContext.contextPath}/sites/${project.unixName}/networks">Explore</a></li>
+<li ${currentPage == "exploregraph" ? "class=\"active\"" : ""}><a href="${pageContext.servletContext.contextPath}/sites/${project.unixName}/statistics">Statistics</a></li>
```
jdamerow, line 4: you should use your own current page value to activate the correct tab when the user is on the stats page.

jdamerow:
please reopen this PR when you're done.

PR #187, opened 2016-04-01
Story/quad 119

ajaymodi:


Quadriga/src/main/java/edu/asu/spring/quadriga/service/publicwebsite/impl/ProjectStats.java
```diff
@@ -0,0 +1,198 @@
...
+            return mapUserWorkspace;
+
+        UserStats userStats = (UserStats) setUserWorkspaceStats(projectId,
+                username);
+        mapUserWorkspace.put(username, userStats);
+
+        return mapUserWorkspace;
+    }
+
+    public IUserStats setUserWorkspaceStats(String projectId, String username)
```
jdamerow, line 136: this is not setting something but returning the requests object, so it shouldn't be called "setUser...". "Set.." implies you're setting a value.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/publicwebsite/impl/ProjectStats.java
```diff
@@ -0,0 +1,198 @@
...
+                mapUserWorkspace.put(username, userStats);
+            }
+        }
+
+        String username = projectDeepMapper.getProjectDetails(projectId)
+                .getOwner().getUserName();
+        if (username == null)
+            return mapUserWorkspace;
+
+        UserStats userStats = (UserStats) setUserWorkspaceStats(projectId,
```
jdamerow, line 129: why do you cast here?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/publicwebsite/impl/ProjectStats.java
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    @Override
+    public Map<String, IUserStats> getCountofNetwork(String projectid)
+            throws QuadrigaStorageException {
+        List<INetwork> networksList = networkManager
+                .getNetworksInProject(projectid);
+
+        if (networksList == null)
+            return null;
```
jdamerow, line 157: I would return an empty map here, since there are effectively no network statistics available for the project.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/publicwebsite/impl/ProjectStats.java
```diff
@@ -0,0 +1,198 @@
...
+        for (INetwork network : networksList) {
+            if (network.getCreator() != null) {
+                String username = network.getCreator().getUserName();
+                IUserStats userStats = mapUserNetworks.get(username);
+                if (userStats == null) {
+                    userStats = new UserStats(username, 0, 1);
+                } else {
+                    userStats.incrementNetworkCount();
+                }
+                mapUserNetworks.put(username, userStats);
```
jdamerow, line 170: you only have to put it in the map, if it isn't in there yet.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/publicwebsite/impl/ProjectStats.java
```diff
@@ -0,0 +1,198 @@
...
+        }
+        return mapUserNetworks;
+    }
+
+    @Override
+    public List<IUserStats> getUserStats(String projectId)
+            throws QuadrigaStorageException {
+
+        Map<String, IUserStats> mapWorkspaceStats = getCountofWorkspace(projectId);
+        Map<String, IUserStats> mapNetworkStats = getCountofNetwork(projectId);
```
jdamerow, line 181: Now, I'm getting confused. Are you calling getCountofNetwork and getCountofWorkspace anywhere outside of this class? If not and if they are not intended to, then these methods should be private. Also, in this case create the map once and hand it over to your private methods to be filled so that you can get rid of the extra looping below. If you intend for other classes to be able to call getCountofNetwork and getCountofWorkspace, and have getUserStats as a helper method to get both stats at once, then I would revise your UserStats class, so that you do not have one value always be 0, but either two separate classes for networks and workspaces (possible combined through a third one that this class returns), or just return maps in getCountofNetwork and getCountofWorkspace. Does this make sense?

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/ProjectStatsController.java
```diff
@@ -0,0 +1,129 @@
...
+        JSONArray jsonArray = new JSONArray();
+
+        for (int i = 0; i < length; i++) {
+            JSONObject jsonObject = new JSONObject();
+            IConceptStats conceptStats = conceptsList.get(i);
+            jsonObject.put("conceptId",
+                    conceptStats.getConceptId().replace("\"", ""));
+            jsonObject.put("description", conceptStats.getDescription()
+                    .replace("\"", ""));
+            jsonObject.put("label", conceptStats.getLemma().replace("\"", ""));
```
jdamerow, line 69: Is there a reason you are deleting all quotation marks from the label and description?

jdamerow:
please reopen this PR when you're done.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/ProjectStatsController.java
ajaymodi, line 69: This was done in story-108 by @bharathsharma . He might have a better idea why he needed that.

bharathsharma, line 69: @ajaymodi @jdamerow
If there was a " in the the label or description, the json object created was turning out to be invalid and as a result JSON.parse(data) used to fail. So to clean string, i had to remove it.

PR #192, opened 2016-04-06
Story/quad 108

ajaymodi:


jdamerow:
please resolve conflicts

ajaymodi:
Entire files are coming in conflicts, does only indentation change for those files or anything else as well?

Does it mean I am using wrong indentation?

jdamerow:
I'm sorry, I did some refactoring and also reformatted the files that I changed. I was hoping that nobody would be actually working on those files, and only realized in the end that you made changes to at least one of them. Another thing, that I was going to mention in class on Friday is that if you are on a Windows machine you should set Git's autocrlf to true (see core.autocrlf here: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration). But you're on Linux, right? So unless you have autocrlf set to input, that shouldn't be an issue.

ajaymodi:
Hmm... I tried to resolve conflicts based on that, let me know if you want me to roll back any changes.

Yes, I am using linux machine and I do not see anything related to autocrlf set in my git config so in that case I will be using default option.

Quadriga/src/test/java/edu/asu/spring/quadriga/service/publicwebsite/impl/ProjectStatsTest.java
```diff
@@ -0,0 +1,105 @@
...
+        List<IConceptStats> conceptStatsList = projectStatsUnderTest
+                .getConceptCount(mockedNetworkList);
+        assertTrue(conceptStatsList.containsAll(mockedConceptStatsList)
+                && mockedConceptStatsList.containsAll(conceptStatsList));
+
+    }
+
+    @Test
+    public void getSortedConceptCountTest() throws QuadrigaStorageException {
+        assertEquals("conceptCount", mockedConceptStatsList,
```
jdamerow: Why do you use a "conceptCount" message here? Either don't put anything, or make it a real message.

jdamerow:
ok to test

Quadriga/src/test/java/edu/asu/spring/quadriga/service/publicwebsite/impl/ProjectStatsTest.java
jdamerow: never mind. I'll fix it real quick.

ajaymodi: Thanks.

PR #195, opened 2016-04-08
Story/quad77

rmanojcit:
network fullscreen changes


Quadriga/src/main/webapp/WEB-INF/views/editing/networkvisual.jsp
```diff
@@ -52,4 +52,90 @@ $("input[type=button]").button().click(function(event) {
...
 
 <div id="log"></div>
+<script type="text/javascript">
+
+  function goFullscreen(id) {
+	  document.getElementById('chart').style='position: fixed; top: 0; bottom :0; left: 0; right :0';
+    // Get the element that we want to take into fullscreen mode
+    var a = document.getElementById('chart');
+    document.getElementsByTagName('svg')[0].id = 'svg_id';
+    var svgDoc= document.getElementById('svg_id');
```
jdamerow: why do you first get the element by its tag name, set the id, and then get it by its id? Why don't you just get it by its tagname right away?

Quadriga/src/main/webapp/WEB-INF/views/editing/networkvisual.jsp
```diff
@@ -34,7 +34,7 @@ $("input[type=button]").button().click(function(event) {
 
 <body
 	onload="d3visualize(<c:out value='${jsonstring}'></c:out>,<c:out value='${networkid}'></c:out>,<c:out value='"${pageContext.servletContext.contextPath}"'></c:out>,'force');" />
-
+<a href="#" onclick="goFullscreen('chart')"> <img src="${pageContext.servletContext.contextPath}/resources/fullscreen_icon.png" width="30" height="30"  style="float: left"/> </a>
```
jdamerow: why don't you use font-awesome instead of an image? Easier to maintain and change.

jdamerow:
As I sad last week in class, if you use a windows machine, please read about autocrlf:
https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

don't commit classpath and settings (you changed the line endings I think)

PR #198, opened 2016-04-08
Story/quad 148

suprajkolluri:


jdamerow:
ok to test

jdamerow:
Let's use the public page layout for the 404 page (the one with the blue navigation). Switching the skeleton is probably enough:
image

PR #206, opened 2016-04-13
Story/quad 149

v1kd:
Please review.


Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformer.java
```diff
@@ -35,45 +28,60 @@
...
+    private EventParser parser;
 
     @Autowired
-    private EventParser parser;
+    private ElementEventTypeDownloadService elementEventTypeDownloadService;
 
     private static final Logger logger = LoggerFactory
             .getLogger(NetworkTransformer.class);
 
+    private static int NUM_THREADS = 4;
```
jdamerow: Let's make this configurable. Depending on the capabilities of the machine, admins might want to adjust this.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformer.java
```diff
@@ -35,45 +28,60 @@
...
-                }
-            }
+        if (networkTopNodesList == null || networkTopNodesList.size() == 0) {
+            // return the trasnformed network
+            return transformedNetwork;
         }
 
+        // networkTopNode list has size > 0
+        // Create executors and tasks
+        ExecutorService executorService = Executors.newFixedThreadPool(NUM_THREADS);
```
jdamerow: Let's use Springs TaskExecutor here and inject it (http://docs.spring.io/spring/docs/current/spring-framework-reference/html/scheduling.html#scheduling-task-executor-usage). If I'm not wrong you create a new thread pool every time the method is called, which kind of goes against the idea of a thread pool.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformer.java
```diff
@@ -35,45 +28,60 @@
...
+                Future<ElementEventsType> future = completionService.take();
+                // take() will block if none are executed
+                ElementEventsType elementEventsType = future.get();
+                parser.parseStatement(networkNodeInfo.getId(), elementEventsType, nodes, links);
+            } catch (InterruptedException | ExecutionException e) {
+                logger.error("Issue while retrieving and marshalling object", e);
+            }
+        }));
+        // finally shutdown the service
+        executorService.shutdown();
```
jdamerow: Also, let's put all of this code in its own class. The class should have a method that returns a lit of ElementEventsType once all threads are done. Then this method here can call the parser for each element. Does this make sense? This way, the parser does the parsing, your new class does the thread/future handling, and this class just coordinates the transformation. single responsibility principal ;)

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformer.java
v1kd: With TaskExecutor I dont think we can use Future or Callable. Is that fine?

v1kd:
Made changes as per your comments.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformer.java
jdamerow: i'm pretty sure you could have used  your previous code, as the taskExecutor implemented the same interface as the executorservice. But looks fine too.

jdamerow:
ok to test

PR #212, opened 2016-04-15
Quad-59 : Fixing bug on delete button of project page.

kavinya:
Quad-59 : Fixing bug on delete button of project page.


jdamerow:
ok to test

PR #220, opened 2016-04-20
Story/quad 86

prasanthpriyanesan:
"Quad-86 Updated the JSP Page according to the comments from previous commit"


Quadriga/src/main/java/edu/asu/spring/quadriga/web/workbench/PublicPageController.java
```diff
@@ -0,0 +1,68 @@
...
+     */
+
+    @RequestMapping(method = RequestMethod.POST, value = "auth/workbench/{projectid}/addpublicpagesuccess")
+    public @ResponseBody ResponseEntity<String> addpublicpagesuccess(
+            @RequestParam("data") String data)
+            throws QuadrigaStorageException, QuadrigaAccessException {
+        if (data.isEmpty())
+            return null;
+        else
+            //Manager class will be called here
```
jdamerow, line 64: as we talked earlier, please write this code here. Work with @TharaSridhar to agree on an interface that you can autowire, and then you should fill this class, as you know best how it works with the jsp page.

Quadriga/src/main/webapp/WEB-INF/views/workbench/addpublicpage.jsp
```diff
@@ -0,0 +1,167 @@
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c"%>
+<%@taglib uri="http://www.springframework.org/tags/form" prefix="form"%>
+
+<!-- Content -->
+<link rel="stylesheet"
+	href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
+<script>
+	function performAction(idValue) {
```
jdamerow, line 8: see previous comments

PR #222, opened 2016-04-20
Story/quad 156

PawanMahalle:


jdamerow:
please resolve conflicts

PawanMahalle:
Resolved conflicts. Please review.

Quadriga/src/main/java/edu/asu/spring/quadriga/dao/impl/projectblog/ProjectBlogEntryDAO.java
```diff
@@ -0,0 +1,74 @@
...
+
+        if (projectId == null || projectId.equals(""))
+            return null;
+
+        // Create a query to get all projects
+        Query query = sessionFactory.getCurrentSession().getNamedQuery("ProjectBlogEntryDTO.findByProjectId");
+
+        // Set max limit of number of items when there is limit on number of
+        // results to fetch
+        if (limit != null) {
```
jdamerow, line 62: this logic should be in your manager class

jdamerow, line 62: business logic = manager class; database logic = dao :)

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
```diff
@@ -0,0 +1,146 @@
...
+     */
+    @AccessPolicies({ @ElementAccessPolicy(type = CheckedElementType.PROJECT, paramIndex = 3, userRole = {
+            RoleNames.ROLE_COLLABORATOR_ADMIN, RoleNames.ROLE_PROJ_COLLABORATOR_ADMIN }) })
+    @RequestMapping(value = "sites/{ProjectUnixName}/addprojectblogentry", method = RequestMethod.GET)
+    public String addProjectBlogEntryForm(@PathVariable("ProjectUnixName") String projectUnixName,
+            @InjectProject(unixNameParameter = "ProjectUnixName") IProject project,
+            @RequestParam("projectId") String projectId, Model model)
+            throws QuadrigaStorageException, QuadrigaAccessException {
+
+        if (project == null) {
```
jdamerow: project will never be null, you can remove this check.

Quadriga/src/main/java/edu/asu/spring/quadriga/dao/impl/projectblog/ProjectBlogEntryDAO.java
jdamerow, line 62: never mind, I misread that. Let's leave it like that for now.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/PublicBlogController.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    /**
+     * reads the number of blog entries to be displayed from user properties
+     * file using {@linkplain Environment} object
+     * 
+     * @return number of blog entries to be displayed specified as
+     *         <code>projectblogentry.list.count<code> property.
+     */
+    private Integer getBlogEntryCount() {
```
jdamerow: but this should be moved to the manager class.

Quadriga/src/main/resources/locale/messages_en_US.properties
```diff
@@ -142,3 +142,6 @@ account_approval.error=Sorry, the account couldn't be approved. There seems to b
 account_approval.account_roles.required=Please specify at least one role for the user.
 account_approval.action.required=Please select "Approve" or "Deny".
 
+# project blog entry error messages
+blog_title.required=Please enter title of blog entry
```
jdamerow: period missing at the end of the sentence.

Quadriga/src/main/resources/locale/messages_en_US.properties
```diff
@@ -142,3 +142,6 @@ account_approval.error=Sorry, the account couldn't be approved. There seems to b
 account_approval.account_roles.required=Please specify at least one role for the user.
 account_approval.action.required=Please select "Approve" or "Deny".
 
+# project blog entry error messages
+blog_title.required=Please enter title of blog entry
+blog_description.required=Please enter description of blog entry
```
jdamerow: should be: "Please enter a text for the new blog entry."

Quadriga/src/test/java/edu/asu/spring/quadriga/mapper/ProjectBlogEntryDTOMapperTest.java
```diff
@@ -0,0 +1,172 @@
...
+    // Mocked variable values
+    public static final String ID = "BLGE_ID_DUMMY";
+    public static final String TITLE = "TITLE_DUMMY";
+    private static final String DESCRIPTION = "DESCRIPTION_DUMMY";
+    private static final String AUTHOR_NAME = "AUTHOR_NAME";
+    private static final Date CREATED_DATE = new Date();
+    private static final String PROJECT_ID = "PROJ_ID_DUMMY";
+
+    @Mock
+    ProjectBlogEntryDTOMapper projectBlogEntryDTOMapper = Mockito.mock(ProjectBlogEntryDTOMapper.class);
```
jdamerow: Why do you mock this?

Quadriga/src/test/java/edu/asu/spring/quadriga/service/impl/projectblog/ProjectBlogEntryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+    /**
+     * tests
+     * {@link IProjectBlogEntryManager#getProjectBlogEntryList(String, Integer)}
+     * method responsible for retrieving list of project blog entries for a
+     * project
+     * 
+     * @throws QuadrigaStorageException
+     */
+    @Test
+    public void Test() throws QuadrigaStorageException {
```
jdamerow: what does this test? bad method name

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/PublicBlogController.java
PawanMahalle: While developing, I considered that controller should be responsible for specifying the number of entries to fetch since the property is for project blog page only. If we move code to specify the number of entries to manager, manager will always fetch 5 entries specified in properties although some other code might expect the ProjectBlogEntryManager to fetch all the blog entries. That's why I added the specified the number of entries to fetch in controller instead of manager.

Could you please confirm if my understanding is correct? Please let me know if I am missing some case here. I am working on rest of the changes suggested by you.

Thanks,
Pawan

Quadriga/src/test/java/edu/asu/spring/quadriga/service/impl/projectblog/ProjectBlogEntryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+        // Assert that exactly 2 entries are fetched
+        assertEquals(projectBlogEntryList.size(), 2);
+
+        // Assert the details of first blog fetched only when atleast one entry
+        // is fetched
+        // If there are fewer entries than expected, first assert will fail and
+        // NullPointerException is avoided
+        if (projectBlogEntryList.size() >= 1) {
+            assertEquals(projectBlogEntryList.get(0).getProjectBlogEntryId(), BLGE_ID_1);
```
jdamerow: assertEquals has first the expected value and then the tested one I believe

Quadriga/src/test/java/edu/asu/spring/quadriga/service/impl/projectblog/ProjectBlogEntryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+    private void assertProjectBlogEntryList(List<IProjectBlogEntry> projectBlogEntryList) {
+
+        // Assert that exactly 2 entries are fetched
+        assertEquals(projectBlogEntryList.size(), 2);
+
+        // Assert the details of first blog fetched only when atleast one entry
+        // is fetched
+        // If there are fewer entries than expected, first assert will fail and
+        // NullPointerException is avoided
+        if (projectBlogEntryList.size() >= 1) {
```
jdamerow: why do you do these if statements here? you already made sure there are two elements in the list.

Quadriga/src/test/java/edu/asu/spring/quadriga/service/impl/projectblog/ProjectBlogEntryManagerTest.java
```diff
@@ -0,0 +1,176 @@
...
+        ProjectBlogEntryDTO projectBlogEntryDTO = createProjectBlogEntryDTO(UNIQUE_BLGE_ID, TITLE_NEW, DESCRIPTION_NEW);
+
+        Mockito.when(projectBlogEntryDTOMapper.getProjectBlogEntryDTO(projectBlogEntry))
+                .thenReturn(projectBlogEntryDTO);
+
+        // Calling the method to test
+        projectBlogEntryManagerUnderTest.addNewProjectBlogEntry(projectBlogEntry);
+
+        // Verifying that saveNewDTO method is called
+        verify(projectBlogEntryDAO).saveNewDTO(projectBlogEntryDTO);
```
jdamerow, line 174: you also want to make sure that the projectBlogEntry that is passed to your mapper has a date and and id.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/PublicBlogController.java
jdamerow: Well, ideally, the the blog would work something like that:
1. there is a default number of blog entries shown on the page. Default should be retrieved in manager class, so that there is no code duplication, if another class uses the same method (to retrieve all blog entries).
2. the blog page might let the user specify that they want to see more or less blog entries on the page (so the controller can call a method in the manager, that let's them specify how many blog entries they have). Basically, controllers control what's coming from the user side, manager control what the user can't influence (e.g. default values).
3. eventually, users want to move trhough the pages of blog entries and we will extend the existing methods to allow specifying pagination. It's even more important for that, that we don't have to calculate the number of entries per page every time.

Does this make sense?

PawanMahalle:
Merged changes from story/QUAD-137 to story/QUAD-156. Please review.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/projectblog/IProjectBlogEntryManager.java
```diff
@@ -0,0 +1,41 @@
...
+ * implementations.
+ * 
+ * @author PawanMahalle
+ *
+ * @see ProjectBlogEntryManager
+ */
+public interface IProjectBlogEntryManager {
+
+    /**
+     * populates the details of a project blog entry and uses DAO to add an
```
jdamerow: You're missing the important things here. The method generates a new ID for the blog entry and it sets its date. Also, remove the table stuff. This is a manager class that doesn't know anything about where things are actually stored.

jdamerow:
ok to test

jdamerow:
Great, looks good. Let's make the font bigger though, to fit with the rest of the website:
image

PR #226, opened 2016-04-22
Quad 138

JNagarkar:


Quadriga/src/main/java/edu/asu/spring/quadriga/dao/impl/transform/TransformFilesDAO.java
```diff
@@ -0,0 +1,43 @@
...
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public Object getDTO(String id) {		
+		return getDTO(TransformFilesDTO.class, id);
+	}
+
+	@SuppressWarnings("unchecked")
+	@Override
+	public void saveTransformDTO(TransformFilesDTO tranformDTO) {
```
jdamerow, line 39: remove. you can just use the method from the parent class.

jdamerow:
see comments on old PR

PR #230, opened 2016-04-24
Story/quad 158

ajaymodi:
First approach. Though working, too distant to be perfect. Require lots of modification mostly in design.


Quadriga/src/main/java/edu/asu/spring/quadriga/dao/workbench/settings/impl/StatisticDAO.java
```diff
@@ -0,0 +1,70 @@
...
+        if (projectStatisticsDTO == null) {
+            projectStatisticsDTO = new ProjectStatisticsDTO();
+            ProjectStatisticsDTOPK psDTOPk = new ProjectStatisticsDTOPK(
+                    projectId, name);
+            projectStatisticsDTO.setProjectStatisticsDTOPK(psDTOPk);
+            projectStatisticsDTO.setIschecked(isChecked);
+            saveNewDTO(projectStatisticsDTO);
+        } else {
+            projectStatisticsDTO.setIschecked(isChecked);
+            updateDTO(projectStatisticsDTO);
```
jdamerow, line 60: This logic should go into your manager class, otherwise if we switch our the backend at any point, we'll have to rewrite this.

Quadriga/src/main/java/edu/asu/spring/quadriga/dao/workbench/settings/impl/StatisticDAO.java
```diff
@@ -0,0 +1,70 @@
...
+            query.setParameter("projectid", projectId);
+            return query.list();
+
+        } catch (HibernateException e) {
+            throw new QuadrigaStorageException(e);
+        }
+    }
+
+    @Override
+    public void upsertStatisticsSettings(String projectId, String name,
```
jdamerow, line 41: method name? Do you mean "update"?

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/factory/impl/workbench/PublicPageFactory.java
```diff
@@ -0,0 +1,44 @@
...
+import edu.asu.spring.quadriga.domain.workbench.IPublicPage;
+
+/**
+ * Factory class to create Public Page object
+ * 
+ * @author Prasanth Priya Nesan
+ * 
+ */
+@Service
+public class PublicPageFactory implements IPublicPageFactory {
```
jdamerow, line 17: why is this class here?

jdamerow, line 17: and the other public page classes?

Quadriga/src/main/java/edu/asu/spring/quadriga/dto/ProjectStatisticsDTO.java
```diff
@@ -0,0 +1,100 @@
...
+@NamedQueries({
+        @NamedQuery(name = "ProjectStatisticsDTO.findAll", query = "SELECT p FROM ProjectStatisticsDTO p"),
+        @NamedQuery(name = "ProjectStatisticsDTO.findByProjectid", query = "SELECT p FROM ProjectStatisticsDTO p WHERE p.projectStatisticsDTOPK.projectid = :projectid"),
+        @NamedQuery(name = "ProjectStatisticsDTO.findByName", query = "SELECT p FROM ProjectStatisticsDTO p WHERE p.projectStatisticsDTOPK.projectid = :projectid and p.projectStatisticsDTOPK.name = :name"), })
+public class ProjectStatisticsDTO implements Serializable {
+
+    private static final long serialVersionUID = 1L;
+
+    @EmbeddedId
+    protected ProjectStatisticsDTOPK projectStatisticsDTOPK;
```
jdamerow, line 41: I'd just put projectId and name properties in this class and give it an id. Makes your code much simpler.

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/factory/impl/workbench/PublicPageFactory.java
ajaymodi, line 17: This class was taken from story 86, as I have to refer previous story to comply with it.

Quadriga/src/main/java/edu/asu/spring/quadriga/dao/workbench/settings/impl/StatisticDAO.java
ajaymodi, line 41: upsert is update + insert 

Generally mysql does that if row exists then update, if not insert.

jdamerow, line 41: In hibernate that's called "saveOrUpdate". If you move the logic out into the manager, you just need to to call this method (Rajat created a method for that in the BaseDAO, let me check if it's merged).

Quadriga/src/main/java/edu/asu/spring/quadriga/dto/ProjectStatisticsDTO.java
ajaymodi, line 41: I am sorry, I did not get it. Does it mean I should remove primary key, foreign key concept over here?

jdamerow, line 41: This should look like this:

```
@Id
private String id;

// no constraints on the db here, should be in your business logic
@Column(name = "name") // this is an optional annotation I believe; 
private String name;

// other properties, e.g. projectId.
```

Quadriga/src/main/java/edu/asu/spring/quadriga/dao/workbench/settings/impl/StatisticDAO.java
jdamerow, line 41: Ok, it's not in the develop yet. Just create your own method that calls:

```
sessionFactory.getCurrentSession().saveOrUpdate(yourDTO);
```

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/factory/impl/workbench/PublicPageFactory.java
jdamerow, line 17: Why? All you need is the public page settings link and main page, right? Only take those and not the rest.

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/workbench/PublicStatisticsPage.java
```diff
@@ -0,0 +1,58 @@
...
+import edu.asu.spring.quadriga.domain.workbench.IPublicPage;
+
+/**
+ * @description : Public Statistics Page class describing the properties of a
+ *              settings for Statisctics Page object
+ * 
+ * @author : Ajay Modi
+ * 
+ */
+public class PublicStatisticsPage implements IPublicPage {
```
jdamerow, line 12: Why do you implement IPublicPage?

jdamerow, line 12: You do not need to do anything publicPages. That's just for the text blocks on the main page of the public page part. You need to create your own objects.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,118 @@
...
+ *
+ */
+@Service
+public class StatisticsSettingsService {
+
+    @Autowired
+    private IStatisticDAO statisticDao;
+
+    @Transactional
+    public void updateRow(String projectId, String name, Boolean isChecked)
```
jdamerow, line 32: why is this called "updateRow"

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,118 @@
...
+            throws QuadrigaStorageException {
+        statisticDao.upsertStatisticsSettings(projectId, name, isChecked);
+    }
+
+    @Transactional
+    public void updateSettings(String projectId, String[] names)
```
jdamerow, line 38: This is not an ideal design. If we have more settings for the diagrams in the future (e.g. bar chart diagram color), how would you do that? You would have to change all your classes. This method should handle objects that represent the settings for your diagrams. In other words, this method should get one (or several) objects, each describing one diagram. Right now, the only settings in those objects would be if it's visible or not. Then this method should simply update the corresponding objects in the db.

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/factory/impl/workbench/PublicPageFactory.java
ajaymodi, line 17: My thinking was since it's a part of arranging settings for public page and since we have public page factory - I thought they are linked, as if using factory pattern.

But, okay I will do it the way which you described.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,118 @@
...
+                    updateRow(projectId, name, true);
+                } else {
+                    updateRow(projectId, name, false);
+                }
+            }
+        }
+    }
+
+    @Transactional
+    public List<ProjectStatisticsDTO> getStatisticsSettingsDTOList(
```
jdamerow, line 54: Your manager should not pass on DTOs. DTOs or between manager and dao class. You would either create a domain object (and its interface) that this method passes on and map your DTO to the domain object, or, if your DTO and domain object look the same, you could just create the interface for your domain object and have the DTO implement that. Then you can pass on the DTO, but controllers won't know, because your return type is the interface. This way, if our domain objects ever need to have more functionality than the DTOs (e.g. a method that calculates the best text color based on the chart color), then we can simply add the domain class/mapping part to the manager without having to change the structure of our classes. Does this make sense?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,118 @@
...
+            for (ProjectStatisticsDTO psDTO : statisticsDTOList) {
+                String name = psDTO.getProjectStatisticsDTOPK().getName();
+                names.add(name);
+            }
+        }
+        return names;
+    }
+
+    @Transactional
+    public List<String> getStatisticsSettingsList(String projectId)
```
jdamerow, line 74: Here you should return a list of your settings object (see comment above).

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,118 @@
...
+
+    @Transactional
+    public List<ProjectStatisticsDTO> getStatisticsSettingsDTOList(
+            String projectId) throws QuadrigaStorageException {
+        List<ProjectStatisticsDTO> statisticsDTOList = statisticDao
+                .getStatisticsSettings(projectId);
+        return statisticsDTOList;
+    }
+
+    public List<String> getNames(List<ProjectStatisticsDTO> statisticsDTOList)
```
jdamerow, line 61: according to the next comment, you can then delete this method. In any case it shouldn't be public unless you want other classes to use it.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,118 @@
...
+        List<ProjectStatisticsDTO> statisticsDTOList = statisticDao
+                .getStatisticsSettings(projectId);
+        if (statisticsDTOList == null || statisticsDTOList.size() == 0) {
+            return new ArrayList<String>(getdefaultMapSettings().keySet());
+        }
+        List<String> names = getNames(statisticsDTOList);
+        return names;
+    }
+
+    public Map<String, Boolean> convertIntoMap(
```
jdamerow, line 85: this will go too if you return objects instead of lists and maps

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,118 @@
...
+            for (ProjectStatisticsDTO psDTO : statisticsDTOList) {
+                String name = psDTO.getProjectStatisticsDTOPK().getName();
+                Boolean isChecked = psDTO.getIschecked();
+                statisticsSettings.put(name, isChecked);
+            }
+        }
+        return statisticsSettings;
+    }
+
+    public Map<String, Boolean> getdefaultMapSettings() {
```
jdamerow, line 99: That's a good example why you would use objects. This is counter-intuitive and highly error prone for maintaining. What does this mean that "concept statics" is false? That concept statistics are not shown by default? And to you have hard coded strings everywhere to pass those values around? That will definitely lead to errors down the road as just a simple typo can break things.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,118 @@
...
+        Map<String, Boolean> statisticsSettings = new HashMap<String, Boolean>();
+        statisticsSettings.put("concept statistics", false);
+        statisticsSettings.put("user statistics", false);
+        return statisticsSettings;
+    }
+
+    @Transactional
+    public Map<String, Boolean> getStatisticsSettingsMap(String projectId)
```
jdamerow, line 107: return objects here

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/factory/impl/workbench/PublicPageFactory.java
jdamerow, line 17: Ya, that was a misunderstanding. The public page objects are just for the text blocks on the main page. Good example, how important good javadoc is ;)

Quadriga/src/main/webapp/WEB-INF/views/workbench/public/publicpagestatistics.jsp~
```diff
@@ -0,0 +1,151 @@
+<%@ taglib uri="http://java.sun.com/jsp/jstl/core" prefix="c"%>
+<%@taglib uri="http://www.springframework.org/tags/form" prefix="form"%>
```
jdamerow, line 2: Looks like you checked in a temporary file.

Quadriga/src/main/webapp/WEB-INF/views/workbench/public/publicpagestatistics.jsp
```diff
@@ -0,0 +1,66 @@
...
+
+			$("form input:checkbox").prop("checked", false);
+			event.preventDefault();
+			return;
+
+		});
+
+	});
+</script>
+<input type="button" value="Select All" name="selectAll">
```
jdamerow, line 31: How does the user know what they have to do on this page? Some more info what the user can do here is needed.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workbench/PublicPageController.java
```diff
@@ -0,0 +1,132 @@
...
+     * page on publicly accessible page
+     *
+     * @author Ajay Modi
+     * @return model - model object
+     */
+
+    @AccessPolicies({ @ElementAccessPolicy(type = CheckedElementType.PROJECT, paramIndex = 1, userRole = {
+            RoleNames.ROLE_COLLABORATOR_ADMIN,
+            RoleNames.ROLE_PROJ_COLLABORATOR_ADMIN }) })
+    @RequestMapping(value = "auth/workbench/{projectid}/loadpublicstatistics", method = RequestMethod.GET)
```
jdamerow, line 93: create your own controller. One controller per page in most cases.

jdamerow, line 93: also, let's choose a shorter page name. maybe just "statistics" (instead of "loadpublicstatistics")?

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workbench/PublicPageController.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    /**
+     * This method is used update the database with the information provided in
+     * the Public Statistics settings page
+     *
+     * @return json
+     */
+
+    @RequestMapping(method = RequestMethod.POST, value = "auth/workbench/{projectid}/submitpublicstatistics")
```
jdamerow, line 118: your methods need an accesspolicy. only project owners and admins should be allowed to change settings.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workbench/PublicPageController.java
```diff
@@ -0,0 +1,132 @@
...
+    public @ResponseBody ResponseEntity<String> submitStatisticsSettings(
+            @ModelAttribute("SpringWeb") PublicStatisticsPage publicStatisticsPage,
```
jdamerow, line 120: don't call this "SpringWeb" just use the default name or a more appropriate one.

jdamerow, line 120: why do you return a responseBody here? Unless you plan on using Ajax to save your page, which isn't really necessary as you only have one form to submit, you can just use a regular modelAndView.

Quadriga/src/main/java/edu/asu/spring/quadriga/dto/ProjectStatisticsDTO.java
ajaymodi, line 41: I am representing (projectId, name) as composite primary key, where projectId remains dependent on projects table. So all this constrains should be in business logic? Do I have to add validator which checks it?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
ajaymodi, line 38: So in that case, settings object is DTO?, hmm... next comment says manager should not pass DTOs then what is my settings object exactly? and why does it have dependency on form tags?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
ajaymodi, line 99: So, does default values should set in properties?
Should I make a new object?
or should I store in database in default table?

Quadriga/src/main/webapp/WEB-INF/views/workbench/public/publicpagestatistics.jsp~
ajaymodi, line 2: Yeah, should not do git commit in hurry.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workbench/PublicPageController.java
ajaymodi, line 120: Yeah, should not do copy paste. Too error prone.

So, does it mean, modelandview only works with single model and view combination? Is it to represent tight bound between model and view?

In that case, just whatever pass from form would be in the formatted manner, so I can cut down all the logic from service and just pass over object to DAO to store.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
ajaymodi, line 54: Finally, thats where magic lies...

PR #241, opened 2016-04-28
Story/quad 158

ajaymodi:
Except two issues, I have address all comments.
1. Since addpublicpage handles by publicpagecontroller and publicpagefactory was require inside method of the same controller, I kept those classes in my code 
2. while adding access policy in post request in submitting form, I am getting null exception by checkauthorization method, which I am unable to resolved.

Further, I am also making two assumptions 
1. best practice to load default values are from resource files
2. making composite primary key of (projectid, settingsname) will help to handle database operation better way rather than making individual column.

Thanks.


Quadriga/src/main/java/edu/asu/spring/quadriga/dao/workbench/settings/impl/StatisticSettingsDAO.java
```diff
@@ -0,0 +1,70 @@
...
+            return SUCCESS;
+        } catch (Exception e) {
+            logger.error("saveOrUpdateStatisticsSettings method :", e);
+            throw new QuadrigaStorageException(e);
+        }
+    }
+
+    @Override
+    public StatisticsSettingsDTO getDTO(String id) {
+        return null;
```
jdamerow, line 67: you need to implement this method

jdamerow:
"making composite primary key of (projectid, settingsname) will help to handle database operation better way rather than making individual column." Why? Do you have a specific situation in mind?
I believe a composite PK like this will make it more difficult to adjust in the future. If we for example decide users can choose their own settings, then this primary key wouldn't work any longer.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,89 @@
...
+                        name, true);
+            } else {
+                statisticsSettingsDTO = new StatisticsSettingsDTO(projectId,
+                        name, false);
+            }
+
+            statisticSettingsDao
+                    .saveOrUpdateStatisticsSettings(statisticsSettingsDTO);
+        }
+    }
```
jdamerow, line 54: Ok, I see why you're doing what you did. But you could simply write a query that gets the object by its projectid and name and only need to change your code slightly.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,89 @@
...
+        List<StatisticsSettingsDTO> statisticsDTOList = null;
+        String[] defaultValues = getDefaultStatisticsSettings();
+        statisticsDTOList = statisticSettingsDao
+                .getStatisticsSettings(projectId);
+        if (statisticsDTOList.size() == 0) {
+            statisticsDTOList = new ArrayList<StatisticsSettingsDTO>();
+            for (String name : defaultValues) {
+                StatisticsSettingsDTO statisticsSettingsDTO = new StatisticsSettingsDTO(
+                        projectId, name, false);
+                statisticsDTOList.add(statisticsSettingsDTO);
```
jdamerow, line 68: here you would just get all dtos by project id.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,89 @@
...
+        for (StatisticsSettingsDTO statisticsSettingsDTO : statisticsDTOList) {
+            IStatisticsSettings s = StatisticsSettingsDTOMapper
+                    .getStatisticsSettings(statisticsSettingsDTO);
+            statisticsSettingsList.add(s);
+        }
+
+        return statisticsSettingsList;
+    }
+
+    private String[] getDefaultStatisticsSettings() {
```
jdamerow, line 82: that's a bad method name. Sounds like it would return default values, but it just returns the different setting types.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
```diff
@@ -0,0 +1,89 @@
...
+
+        String propertyValue = env.getProperty("statistics.settings");
+        String[] settings = propertyValue.split(",");
+        return settings;
```
jdamerow, line 86: I'd return a list here, as it is pretty likely you'll more often need a list than an array.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workbench/StatisticsSettingsController.java
```diff
@@ -0,0 +1,93 @@
...
+import edu.asu.spring.quadriga.web.login.RoleNames;
+
+@Controller
+public class StatisticsSettingsController {
+
+    @Autowired
+    private StatisticsSettingsService statsSettingsService;
+
+    @Autowired
+    private StatisticsSettingsBean statisticsSettingsBean;
```
jdamerow, line 31: don't autowire this.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workbench/StatisticsSettingsController.java
```diff
@@ -0,0 +1,93 @@
...
+            @PathVariable("projectid") String projectid)
+            throws QuadrigaStorageException, QuadrigaAccessException {
+
+        ModelAndView model = new ModelAndView("auth/workbench/publicstatistics");
+        model.getModelMap().put("publicpageprojectid", projectid);
+        List<IStatisticsSettings> statisticsSettingsList = statsSettingsService
+                .getStatisticsSettingsList(projectid);
+        model.getModelMap().put("statistics", statisticsSettingsList);
+        model.getModelMap().put("statisticsSettingsBean",
+                statisticsSettingsBean);
```
jdamerow, line 55: just use new keyword here

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workbench/StatisticsSettingsController.java
```diff
@@ -0,0 +1,93 @@
...
+        ModelAndView model = new ModelAndView("redirect:/auth/workbench/"
+                + projectid + "/statistics");
+        try {
+            String[] names = statisticsSettingsBean.getNames();
+            statsSettingsService
+                    .addOrUpdateStatisticsSettings(projectid, names);
+            attr.addFlashAttribute("show_success_alert", true);
+            attr.addFlashAttribute("success_alert_msg",
+                    "Settings has been updated successfully.");
+        } catch (Exception e) {
```
jdamerow, line 82: don't catch "exception" be specific

Quadriga/src/main/webapp/WEB-INF/views/workbench/public/publicpagestatistics.jsp
```diff
@@ -0,0 +1,70 @@
...
+					<c:choose>
+						<c:when test="${statistic.isChecked}">
+							<td><form:checkbox path="names" value="${statistic.name}"
+									checked="checked" /></td>
+						</c:when>
+						<c:otherwise>
+							<td><form:checkbox path="names" value="${statistic.name}" /></td>
+						</c:otherwise>
+					</c:choose>
+					<td><font size="3"> <c:out value="${statistic.name}"></c:out>
```
jdamerow, line 62: This will result in the page displaying 'user_statistics' next to the checkbox, right? This should be real text like "User statistics". You can just use the key in a message file to assign a real label to each key. Does this make sense?

jdamerow:
Please remove all the public page files as that will create conflicts with the branch that actually contains them.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/workbench/StatisticsSettingsService.java
ajaymodi, line 68: I believe I am taking all dtos by projectID, however if there is no entry in database yet, then I am iterating over default values.

PR #244, opened 2016-04-28
Quad 138

JNagarkar:
Added the networks based on projects instead of users. Necessary changes made.


JNagarkar:
https://github.com/diging/quadriga/pull/237
Reopened the above request and close this one.

PR #260, opened 2016-05-09
Quad 160

nischalsamji:
Add a Rest Endpoint for Uploading Texts to Quadriga.


Quadriga/src/main/java/edu/asu/spring/quadriga/rest/TextUploadRestController.java
```diff
@@ -0,0 +1,134 @@
...
+            txtFile = txtXMLParser.parseTextXML(xml, wsId, projId);
+        } catch (TextFileParseException tfe) {
+            logger.error("Error in Text Rest Controller:" + tfe);
+            String errorMsg = errorMessageRest.getErrorMsg(tfe.getMessage());
+            return new ResponseEntity<String>(errorMsg, HttpStatus.BAD_REQUEST);
+        }
+        try {
+            tfManager.saveTextFile(txtFile);
+        } catch (QuadrigaStorageException e) {
+            logger.error("Error in Text Rest Controller:" + e);
```
jdamerow: that should be 

```
logger.error("msg", e)
```

jdamerow: for proper display

Quadriga/src/main/java/edu/asu/spring/quadriga/utilities/impl/TextXMLParser.java
```diff
@@ -0,0 +1,93 @@
...
+
+        DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+
+        DocumentBuilder builder;
+        Document document = null;
+        try {
+            builder = factory.newDocumentBuilder();
+            document = builder.parse(new InputSource(new StringReader(xml)));
+        } catch (ParserConfigurationException | IOException e) {
+            logger.error("Error while parsing the XML", e);
```
jdamerow: don't log the whole stack trace here. You can log a message but assume that the exception will be handled above (meaning someone will do the correct thing when an exception is thrown which also includes logging). Otherwise, there will be two stacktraces printed for the same exception. Does this make sense?

Quadriga/src/main/java/edu/asu/spring/quadriga/utilities/impl/TextXMLParser.java
```diff
@@ -0,0 +1,93 @@
...
+            throw new TextFileParseException("Error while parsing the XML");
```
jdamerow: you still need to pass the exception to the new exception or you loose the "caused by..." stacktrace.

jdamerow:
almost there

jdamerow:
@nischalsamji please reoslve conflicts

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/TextUploadRestController.java
```diff
@@ -0,0 +1,134 @@
...
+            engine = restVelocityFactory.getVelocityEngine(request);
+            engine.init();
+            Template template = engine.getTemplate("velocitytemplates/textfile.vm");
+            VelocityContext context = new VelocityContext(restVelocityFactory.getVelocityContext());
+            context.put("textid", txtFile.getTextId());
+            context.put("refid", txtFile.getRefId());
+            context.put("filename", txtFile.getFileName());
+            context.put("wsid", txtFile.getWorkspaceId());
+            context.put("projid", txtFile.getProjectId());
+            context.put("texturi", textURI + "/" + txtFile.getTextId());
```
jdamerow: textURI should be a property of a textfile.So, here you should simply call a method on the text file to get its URI. The manager class handling text files should be the one who knows how to create the URIs. Otherwise, you'll have to change every piece of code where you create a uri if we change its structure or anything.

Quadriga/src/main/resources/locale/projectconstants.properties
```diff
@@ -1,5 +1,6 @@
 #Project Unix name URL
 project_unix_name.url = http://quadriga.asu.edu/sites/
+textfiles_location.url = http://quadriga.asu.edu/textfiles
```
jdamerow: should should just be `textfiles.uri`
and for the value, let's use `http://quadriga.io/text/`

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/TextUploadRestController.java
nischalsamji: The URI here I am setting is for the public pages/ texts[quadriga.io/text/]. Should this be part of the TextFile ?

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/workspace/TextFile.java
```diff
@@ -16,6 +16,7 @@
     private String textId;
     private String fileContent;
     private String refId;
+    private String TextFileURI;
```
jdamerow: This is a variable. Don't capitalize variable names!

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/TextUploadRestController.java
```diff
@@ -0,0 +1,130 @@
...
+ * 
+ */
+@Controller
+public class TextUploadRestController {
+
+    @Autowired
+    ITextXMLParser txtXMLParser;
+
+    @Autowired
+    ITextFileManager tfManager;
```
jdamerow: unless there is a good reason, your variables should be private

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/workspace/TextFile.java
```diff
@@ -72,4 +73,12 @@ public void setRefId(String refId) {
         this.refId = refId;
     }
 
+    public String getTextFileURI() {
+        return TextFileURI;
+    }
+
+    public void setTextFileURI(String textFileURI) {
```
jdamerow: This is not a good parameter name. It implies that I give the final file URI to the method, but that's not what the method needs. Especially with method names like this (getter/setter), you need to make very clear what the method expects. Othrewise, I assume that the default behavior applies. Also, make sure to write javadoc for this method in which you explain what it does!

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/workspace/TextFile.java
```diff
@@ -72,4 +73,12 @@ public void setRefId(String refId) {
         this.refId = refId;
     }
 
+    public String getTextFileURI() {
+        return textFileURI;
+    }
+
+    public void setTextFileURI(String URIPrefix) {
```
jdamerow: same here, don't start variable/parameter names with an uppercase letter

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/workspace/ITextFile.java
```diff
@@ -29,5 +29,14 @@
     public abstract String getFileContent();
 
     public abstract void setFileContent(String fileContent);
+    
+    public abstract String getTextFileURI();
+
+    /**
+     * @param URIPrefix
+     * 
+     * Sets the URI corresponding to the TextFile.
```
jdamerow: I can guess that from the method name. Your javadoc should tell me what the method does and what it expects. Also, please review how to structure javadoc.

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/workspace/ITextFile.java
```diff
@@ -29,5 +29,16 @@
...
+    
+    public abstract String getTextFileURI();
+
+    /**
+     * @param uriPrefix
+     *      Prefix of the URI specified in the properties file.
+     *      eg: http://quadriga.io/text/textid
+     * 
+     * Sets the URI corresponding to the TextFile. 
+     */
```
jdamerow, line 42: Read this:
http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#format
and this:
http://www.monperrus.net/martin/how-to-write-good-API-documentation

And then rewrite your javadoc accordingly.

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/workspace/ITextFile.java
```diff
@@ -29,15 +29,16 @@
     public abstract String getFileContent();
 
     public abstract void setFileContent(String fileContent);
-    
+
     public abstract String getTextFileURI();
 
     /**
+     * This method sets the uri of the related TextFile based on the passed to
```
jdamerow: "based on the passed" what? Also, what does "related TextFile" mean? Important here is that this method uses the id of the file as part of the URI.

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/workspace/ITextFile.java
```diff
@@ -29,15 +29,16 @@
...
+     * the method.
+     * 
      * @param uriPrefix
-     *      Prefix of the URI specified in the properties file.
-     *      eg: http://quadriga.io/text/textid
+     *            Prefix of the URI specified in the properties file.
```
jdamerow: this method doesn't care where the prefix is coming from. All it cares about is that it gets the base URI/prefix for generating uris.

jdamerow:
ok to test

jdamerow:
@nischalsamji please check the test cases. looks like something is broken.

nischalsamji:
Updated Test Case

jdamerow:
Hm, when I submit a text that contains special characters it'll give an XML parse exception. When I encode the text before submitting it works, but when I click on the text from the workspace page, it'll show me the encoded text. Somewhere in between you need to decode. Try with the text from here:
http://www.blindtextgenerator.com/lorem-ipsum
(with dummy text "a-z A-Z 123")

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workspace/ViewTextController.java
```diff
@@ -42,7 +43,9 @@
         String respMessage = null;
         try {
             respMessage = tfManager.retrieveTextFileContent(txtId);
-            return new ResponseEntity<String>(respMessage, HttpStatus.OK);
+            String decodedRespMsg = StringEscapeUtils.unescapeXml(respMessage);
```
jdamerow: Is the text still escaped when stored? It needs to be unescaped before storing, otherwise if someone submits an XML file, it won't be an actual XML file that is stored, which would be bad for reusability purposes.

nischalsamji: Yea, changed code to store un escaped XML. But if the user sends un escaped xml via REST, it will still cause an exception.

PR #275, opened 2016-06-02
[QUAD-180] adjusted the layout in the main page settings by using

yyk16293:
bootstrap classes for input fields, commenting out statistics page,
adding new save icon.

### DIFF
--- a/Quadriga/src/main/webapp/WEB-INF/views/navigation/nav-settings.jsp
+++ b/Quadriga/src/main/webapp/WEB-INF/views/navigation/nav-settings.jsp
@@ -8,27 +8,31 @@
 <script
 	src="${pageContext.servletContext.contextPath}/resources/txt-layout/js/jstree.min.js"></script>
 
-
 <div>
-    <ul>
-        <li><a
-            href="${pageContext.servletContext.contextPath}/auth/workbench/projects/${project.projectId}"><span
-                class="glyphicon glyphicon-circle-arrow-left"></span> Back to Project
-        </a></li>
-    </ul>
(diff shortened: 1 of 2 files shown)


Quadriga/src/main/webapp/WEB-INF/views/settings/addpublicpage.jsp
```diff
@@ -103,43 +103,42 @@
 
 				<tr>
 					<td style="width: 170px">Title *</td>
-					<td style="width: 400px"><input path="title" size="60"
-						class="titleValue" id="title1" value="${publicpageObject0.title}" /></td>
-					<td><div id="title_err"></div></td>
-					<td><form:errors path="title" class="ui-state-error-text"></form:errors></td>
+					
+					<td style="width: 600px"><input path="title" size="60"
```
jdamerow, line 107: Didn't we talk about this? the column shouldn't have any width. Bootstrap should take the complete window width on its own.

Quadriga/src/main/webapp/WEB-INF/views/navigation/nav-settings.jsp
```diff
@@ -8,27 +8,31 @@
...
+<br/>
+
+<div class="panel panel-default">
+  
+    <div class="panel-heading">
+    <h3 class="panel-title">Settings</h3>
+    </div>
+    
+    <div class="list-group">
+    <a class="list-group-item"
```
jdamerow, line 25: the items are never active, are they? Depending on which page a user is, the corresponding menu entry should be active.

PR #282, opened 2016-06-03
Bug/quad 186

nischalsamji:
Redirect to error page when there is an exception in concept Controller.

### DIFF
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/web/conceptcollection/ConceptcollectionController.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/web/conceptcollection/ConceptcollectionController.java
@@ -11,6 +11,8 @@ import java.util.Set;
 import javax.servlet.http.HttpServletRequest;
 
 import org.codehaus.jettison.json.JSONException;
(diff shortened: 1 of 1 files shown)


Quadriga/src/main/java/edu/asu/spring/quadriga/web/conceptcollection/ConceptcollectionController.java
```diff
@@ -161,18 +165,24 @@ private void fillModel(String collectionId, ModelMap model, String username) thr
...
+                    @Override
+                    public int compare(ConceptEntry o1, ConceptEntry o2) {
+                        return o1.getLemma().toLowerCase().compareTo(o2.getLemma().toLowerCase());
+                    }
+                });
+                model.addAttribute("result", conReply.getConceptEntry());
+            }
+        } catch (HttpMessageNotReadableException hex) {
+            logger.error("error:", hex);
+            return "auth/storageissue";
```
jdamerow: if you just throw a Quadriga exception the general exception handler should take care of displaying it. this makes sense if you actually do something in response to it.

jdamerow:
ok to test

PR #283, opened 2016-06-03
[QUAD-188]: Fixed Navigation issue in Worksapce Details page

nischalsamji:
Fixed UI bug that causes Invalid options to pop up.


jdamerow:
ok to test

PR #287, opened 2016-06-08
Story/quad 180

yyk16293:


Quadriga/src/main/webapp/WEB-INF/views/settings/addpublicpage.jsp
```diff
@@ -6,24 +6,30 @@
 	href="https://maxcdn.bootstrapcdn.com/font-awesome/4.5.0/css/font-awesome.min.css">
 <script>
 //@ sourceURL=filename.js
+	
+	$(document).ready(function(){
+	$("#mainPageTab").attr('class','list-group-item');
+	$("#aboutTab").attr('class','list-group-item disabled');
+    });
```
jdamerow, line 13: Use tiles attributes instead of javascript. take a look at how the main navigation is set active and do it similarly.

Quadriga/src/main/webapp/WEB-INF/views/settings/publicWebsiteEditAbout.jsp
```diff
@@ -5,6 +5,11 @@
 <script src="//cdn.tinymce.com/4/tinymce.min.js"></script>
 <script>
 
+$(document).ready(function(){
+	$("#mainPageTab").attr('class','list-group-item disabled');
+	$("#aboutTab").attr('class','list-group-item');
+});
+
```
jdamerow, line 12: see previous comment

PR #289, opened 2016-06-10
Story/quad 180

yyk16293:


Quadriga/src/main/webapp/WEB-INF/tiles-defs.xml
```diff
@@ -611,6 +611,7 @@
 	<definition name="auth/workbench/addpublicpage" extends="base.definition.workbench.nav">
 		<put-attribute name="title"
 			value="Quadriga Workbench - Add Public Page Settings" />
+		<put-attribute name="currentPage" value="settings" />
```
jdamerow, line 614: you have to keep the currentPage from before, otherwise the "Projects" tab won't be highlighted anymore.

Quadriga/src/main/webapp/WEB-INF/views/settings/addpublicpage.jsp
```diff
@@ -6,24 +6,26 @@
...
-		var title = $(obj).closest('div.publicpageform').find('.titleValue').val();
-		var desc = $(obj).closest('div.publicpageform').find('.descValue').val();
-		var order = $(obj).closest('div.publicpageform').find('.orderValue').val();
-		var linkTo = $(obj).closest('div.publicpageform').find('.linkTo').val();
-		var linkText = $(obj).closest('div.publicpageform').find('.linkText').val();
+		var title = $(obj).closest('div.publicpageform').find('#title' + value).val();
+		var desc = $(obj).closest('div.publicpageform').find('#description' + value).val();
+		var order = $(obj).closest('div.publicpageform').find('#orderValue' + value).val();
+		var linkTo = $(obj).closest('div.publicpageform').find('#linkTo' + value).val();
+		var linkText = $(obj).closest('div.publicpageform').find('#linkText' + value).val();
```
jdamerow, line 28: why do you add that value here? it should have worked before, since you look for certain class of child elements.

PR #311, opened 2016-07-18
Story/quad 209

yyk16293:


Quadriga/src/main/java/edu/asu/spring/quadriga/web/resolver/EditResolverController.java
```diff
@@ -0,0 +1,47 @@
...
+    @RequestMapping(value = "/auth/resolvers/update", method = RequestMethod.POST)
+    public String editResolver(Principal principal,
+            @Validated @ModelAttribute("resolver") ProjectHandleResolver resolver, BindingResult results, Model model) {
+
+        if (results.hasErrors()) {
+            model.addAttribute("resolver", resolver);
+            return "auth/resolvers/edit";
+        }
+
+        resolver.setUsername(principal.getName());
```
jdamerow: why do you set the username here?

yyk16293: As it's value was null, I was setting it's value here. I made changes to the code, now it's being handled using hidden value in the form being submitted.

jdamerow: If you submit a hidden field with the correct field name (the one it has in the bean), it'll be set automatically.

jdamerow: never mind. Now I see that you changed it.

jdamerow:
ok to test

PR #319, opened 2016-08-05
[QUAD-220] Making code changes to add lock, unlock buttons for setting

yyk16293:
accessibility of text file.

### DIFF
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/ITextFileManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/ITextFileManager.java
@@ -5,7 +5,6 @@ import java.io.IOException;
 import java.util.List;
 
 import edu.asu.spring.quadriga.domain.workspace.ITextFile;
-import edu.asu.spring.quadriga.dto.TextFileDTO;
 import edu.asu.spring.quadriga.exceptions.FileStorageException;
 import edu.asu.spring.quadriga.exceptions.QuadrigaStorageException;
 
@@ -13,6 +12,7 @@ public interface ITextFileManager {
 
     /**
      * Save the Text file in the database and Local File System.
+     * 
      * @param txtfile
      *            Text File object from web controller to handle text
      *            operations.
@@ -34,12 +34,13 @@ public interface ITextFileManager {
 
     /**
      * Retrieve the text file content as a string.
+     * 
      * @param txtId
      *            Text ID of the file to be retrieved.
      * @return Returns the contents of the file as a string.
      * @throws FileNotFoundException
      * @throws IOException
-     * @throws FileStorageException 
+     * @throws FileStorageException
      */
     String retrieveTextFileContent(String txtId) throws FileStorageException;
 
@@ -48,5 +49,26 @@ public interface ITextFileManager {
     void loadFile(ITextFile txtFile) throws FileStorageException;
 
     ITextFile getTextFile(String textId) throws QuadrigaStorageException;
+
+    /**
+     * saves or updates the txtFile in the database.
+     * 
+     * @param txtFile
+     *            Text File object from web controller to handle text
+     *            operations.
+     * @return
+     */
+    void saveTextFileToDB(ITextFile txtFile);
+
+    /**
+     * If the accessibility of textFile is public, it's set to private.
+     * otherwise it is set to public.
+     * 
+     * @param textFile
+     *            Text File object from web controller to handle text
+     *            operations.
+     * @return textFile with updated accessibility
+     */
+    ITextFile updateTextFileAccessibility(ITextFile textFile);
 
 }
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/ITextFileManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/ITextFileManager.java
@@ -58,7 +58,7 @@ public interface ITextFileManager {
      *            operations.
      * @return
      */
-    void saveTextFileToDB(ITextFile txtFile);
+    void storeTextFile(ITextFile txtFile);
 
     /**
      * If the accessibility of textFile is public, it's set to private.
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/ITextFileManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/ITextFileManager.java
@@ -59,16 +59,4 @@ public interface ITextFileManager {
      * @return
      */
     void storeTextFile(ITextFile txtFile);
-
-    /**
-     * If the accessibility of textFile is public, it's set to private.
-     * otherwise it is set to public.
-     * 
-     * @param textFile
-     *            Text File object from web controller to handle text
-     *            operations.
-     * @return textFile with updated accessibility
-     */
-    ITextFile updateTextFileAccessibility(ITextFile textFile);
-
 }
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/impl/TextFileManager.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/impl/TextFileManager.java
@@ -8,6 +8,7 @@ import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
 import edu.asu.spring.quadriga.dao.textfile.ITextFileDAO;
+import edu.asu.spring.quadriga.domain.enums.ETextAccessibility;
 import edu.asu.spring.quadriga.domain.workbench.IProject;
 import edu.asu.spring.quadriga.domain.workspace.ITextFile;
 import edu.asu.spring.quadriga.dto.TextFileDTO;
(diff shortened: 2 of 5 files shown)


Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/ITextFileManager.java
```diff
@@ -49,4 +50,25 @@
...
 
+    /**
+     * saves or updates the txtFile in the database.
+     * 
+     * @param txtFile
+     *            Text File object from web controller to handle text
+     *            operations.
+     * @return
+     */
+    void saveTextFileToDB(ITextFile txtFile);
```
jdamerow: this is a manager class. method names should not have anything about DBs in them.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/textfile/impl/TextFileManager.java
```diff
@@ -114,7 +120,16 @@ public String retrieveTextFileContent(String txtId) throws FileStorageException
         if (!fileName.contains("."))
             fileName += ".txt";
         return fileSaveServ.retrieveFileFromLocal(fileName, txtId);
-
     }
 
+    @Override
+    public ITextFile updateTextFileAccessibility(ITextFile textFile) {
```
jdamerow: Don't do this here. why don't you just set the accessibility where ever you call this method and then just save it? otherwise, what if we add another accessibility options (e.g. for authorized users only)?

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workspace/UpdateTextAccessibilityController.java
```diff
@@ -0,0 +1,30 @@
...
+    @Autowired
+    private ITextFileManager tfManager;
+
+    @RequestMapping(value = "/auth/workbench/workspace/{workspaceId}/{textId}/updateAccessibility", method = RequestMethod.GET)
+    public String updateTextFileAccessibility(Model model, @PathVariable("textId") String textId,
+            @PathVariable("workspaceId") String workspaceId) throws QuadrigaStorageException, FileStorageException {
+
+        ITextFile textFile = tfManager.getTextFile(textId);
+        ITextFile updatedTextFile = tfManager.updateTextFileAccessibility(textFile);
+        tfManager.saveTextFileToDB(updatedTextFile);
```
jdamerow: here, it make even less sense this way. you tell the manager, change the accessibility and then you make an extra call to save the text file object. why not setting it directly?

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workspace/UpdateTextAccessibilityController.java
```diff
@@ -0,0 +1,30 @@
...
+import edu.asu.spring.quadriga.service.textfile.ITextFileManager;
+
+@Controller
+public class UpdateTextAccessibilityController {
+
+    @Autowired
+    private ITextFileManager tfManager;
+
+    @RequestMapping(value = "/auth/workbench/workspace/{workspaceId}/{textId}/updateAccessibility", method = RequestMethod.GET)
+    public String updateTextFileAccessibility(Model model, @PathVariable("textId") String textId,
```
jdamerow, line 30: don't assume htere are only two types of accessibility. it's an enum field, so we can add an infinite number of options.

Quadriga/src/main/webapp/WEB-INF/views/workspace/workspacedetails.jsp
```diff
@@ -195,9 +195,15 @@
 										<br>
 										<c:if test="${textfile.accessibility == 'PUBLIC'}">
 										<span class="label label-success">Public</span>
+										<a data-toggle="modal" data-target="#changeaccess-ws" id="${textfile.textId}">
+										<i class="fa fa-unlock" aria-hidden="true"></i>
+										</a>
 										</c:if>
 										<c:if test="${textfile.accessibility == 'PRIVATE'}">
                                         <span class="label label-danger">Private</span>
+                                        <a data-toggle="modal" data-target="#changeaccess-ws" id="${textfile.textId}">
```
jdamerow: why is this called `changeaccess-ws`? it's for text files, isn't it?

Quadriga/src/main/webapp/WEB-INF/views/workspace/workspacedetails.jsp
```diff
@@ -382,10 +388,38 @@
...
+		<div class="modal-content">
+			<div class="modal-header">
+				<button type="button" class="close" data-dismiss="modal"
+					aria-label="Close">
+					<span aria-hidden="true">&times;</span>
+				</button>
+				<h4 class="modal-title">Update Text Accessibility</h4>
+			</div>
+			<div class="modal-body">
+				<p>Are you sure you want to change accessibility of text file?</p>
```
jdamerow: This should say: "Are you sure you want to make this text file [public|private]?"  depending on the action, it should say 'public' or 'private'.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workspace/UpdateTextAccessibilityController.java
jdamerow, line 30: this is still here

Quadriga/src/main/webapp/WEB-INF/views/workspace/workspacedetails.jsp
```diff
@@ -401,6 +435,25 @@
...
+							if (textFileAccess === "PUBLIC") {
+								textFileAccess = "Private";
+							} else if (textFileAccess === "PRIVATE") {
+								textFileAccess = "Public";
+							}
+							
+							var content = "Are you sure you want to make this text file " + textFileAccess + "?";
+							$(this).find('.modal-body p').text(content);
+							});
+						});
```
jdamerow: indentation

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workspace/UpdateTextAccessibilityController.java
```diff
@@ -0,0 +1,32 @@
...
+import edu.asu.spring.quadriga.exceptions.QuadrigaStorageException;
+import edu.asu.spring.quadriga.service.textfile.ITextFileManager;
+
+@Controller
+public class UpdateTextAccessibilityController {
+
+    @Autowired
+    private ITextFileManager tfManager;
+
+    @RequestMapping(value = "/auth/workbench/workspace/{workspaceId}/{textId}/{accessStatus}/updateAccessibility", method = RequestMethod.GET)
```
jdamerow: this method changes an object, so it should be post

jdamerow: also, url should end after access status (remove 'updateAccessibility')

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workspace/UpdateTextAccessibilityController.java
```diff
@@ -0,0 +1,32 @@
...
+    public String updateTextFileAccessibility(Model model, @PathVariable("textId") String textId,
+            @PathVariable("workspaceId") String workspaceId, @PathVariable("accessStatus") String accessibility)
+            throws QuadrigaStorageException, FileStorageException {
+
+        ITextFile textFile = tfManager.getTextFile(textId);
+        textFile.setAccessibility(ETextAccessibility.valueOf(accessibility.toUpperCase()));
```
jdamerow, line 50: what if the given textId doesn't exist or the given accessStatus doesnt'?

Quadriga/src/main/webapp/WEB-INF/views/workspace/workspacedetails.jsp
```diff
@@ -401,6 +435,25 @@
...
+						        changedTextFileAccess = "Private";
+						    } else if (textFileAccess === "PRIVATE") {
+						        changedTextFileAccess = "Public";
+						    }
+						    
+						    var content = "Are you sure you want to make this text file " + changedTextFileAccess + "?";
+						    $(this).find('.modal-body p').text(content);
+						});
+	                	});
+</script>
```
jdamerow, line 457: let's format this file

Quadriga/src/main/java/edu/asu/spring/quadriga/web/workspace/UpdateTextAccessibilityController.java
nischalsamji: Can't we use PUT instead of POST as this is an existing object?

jdamerow: by definition, put is usually used to store (or update) an entity at a given url:
"The PUT method requests that the enclosed entity be stored under the supplied Request-URI. If the Request-URI refers to an already existing resource, the enclosed entity SHOULD be considered as a modified version of the one residing on the origin server."

Post basically just says, do something with this:
"The actual function performed by the POST method is determined by the server and is usually dependent on the Request-URI. [...] The action performed by the POST method might not result in a resource that can be identified by a URI. "

If we consider our text file the resource to be updated, then post would be wrong because we're just updating one field here (and we are only sending data for one field). PUT requires that if I send the same response again, I get the same result (in the sense of a resource => text file would be exactly the same). This would not be true because our text file might have changed in between, so although the accessibility would be the same, the text file itself might be different. 

Does this make sense?

nischalsamji: Kind of. Thanks @jdamerow

Quadriga/src/main/resources/locale/messages_en_US.properties
```diff
@@ -129,6 +129,8 @@ workspace.add.not.created=Workspace could not be created.
 workspace.text.add.success = The text file has been successfully saved.
 workspace.text.add.failure = The text file could not be saved.
 workspace.text.add.failure.config = The file could not be saved. If the problem persists, please contact an administrator.
+workspace.text.retrieve.failure = There was no text file with the given textId.
```
jdamerow: should be: "There exists not text file..."

Quadriga/src/main/resources/locale/messages_en_US.properties
```diff
@@ -129,6 +129,8 @@ workspace.add.not.created=Workspace could not be created.
...
+workspace.text.accessibility.failure = There was no accessibility for the text file.
```
jdamerow: should be: "The provided access option does not exist."

jdamerow:
ok to test

jdamerow:
when hovering over the lock icon, it should turn into a hand like other links.

jdamerow:
ok to test

jdamerow:
lock icon/links need tooltip

PR #322, opened 2016-08-16
Story/quad 216

yyk16293:


jdamerow:
ok to test

jdamerow:
The bottom footer should only be as wide as the top navigation bar, with rounded corners. And the contents needs to be properly aligned.
image

jdamerow:
Also, 'Sites' page needs a footer (at /sites).

PR #326, opened 2016-08-26
[QUAD-226] Making code changes to reset modal content once it is hidden

yyk16293:


Quadriga/src/main/webapp/WEB-INF/views/workspace/workspacedetails.jsp
```diff
@@ -467,8 +473,12 @@ cursor: pointer;
                                 changedTextFileAccess = "Public";
                             }
                             
-                            var content = "Are you sure you want to make this text file " + changedTextFileAccess + "?";
-                            $(this).find('.modal-body p').text(content);
+                            var titlecontent = "Update Text Accessibility"
+                            var bodycontent = "Are you sure you want to make this text file " + changedTextFileAccess + "?";
+                            
+                            $(this).find('.modal-title').html(titlecontent);
```
jdamerow, line 479: This will not find the dialog you've created. It'll most likely take the first element with this class it finds.

Quadriga/src/main/webapp/WEB-INF/views/workspace/workspacedetails.jsp
```diff
@@ -467,8 +473,12 @@ cursor: pointer;
                                 changedTextFileAccess = "Public";
                             }
                             
-                            var content = "Are you sure you want to make this text file " + changedTextFileAccess + "?";
```
jdamerow, line 470: Regarding the if-statement before: simply create a map {textFileAccess -> string that should be displayed}. This way you can simply get the string you need and  it's easily extendable.

(I don't seem to have looked too closely at this file before..., sorry).

PR #327, opened 2016-08-29
[QUAD-224]: Making changes to format X-Axis ticks on Project Statistics

nischalsamji:


nischalsamji:
The X-Axis looks like this after this change
screen shot 2016-08-29 at 8 43 27 am

jdamerow:
hm, it all says may 2016. Are those different days in the month? If so, it needs the day as well, otherwise that doesn't make much sense...

nischalsamji:
yup, I changed to include the dae as well. Looks like this now
screen shot 2016-08-29 at 10 29 47 am

jdamerow:
ok to test

jdamerow:
since we're already on this, would you please change the diagram type. the way it is doesn't make much sense. e.g. in the above diagram, how many networks were added on May 21? Can you simply make it a line graph (e.g. like this: http://image.mathcaptain.com/cms/images/113/line-graph.png) please.

nischalsamji:
Changed it to line graph, need some network to test it from QStore...
looks like this
screen shot 2016-08-29 at 1 09 22 pm

jdamerow:
it still has the same issue though: how many networks were submitted on May 21? When does the count go to 0? Shouldn't that be May 13?

nischalsamji:
@jdamerow Looks like this now
screen shot 2016-09-06 at 11 22 48 am

jdamerow:
Hm, something is off in your diagram. It starts at Sep. 11 and goes to Nov. 13 this year. How can that be? How can we plot the future? 

Regarding the diagram itself. Can you just make it a bar chart? I think that makes more sense. Otherwise the graph would say that e.g. Sep. 13 there are 3 or 4 submitted networks although that might just be the connecting line between two points. Also, please simply make the time frame the past month (or 4 weeks).

nischalsamji:
I changed the dates and uploaded few networks to test how the diagram looks.... I will change the line graph to a bar graph.

nischalsamji:
The graph looks like this with the existing data, as you mentioned if I change the time frame, there are blanks coming up in the graph.
screen shot 2016-09-20 at 3 17 55 pm

jdamerow:
That's ok. There can be blanks and there should be blank if nothing has been submitted.

nischalsamji:
Looks like this now
screen shot 2016-09-21 at 8 53 16 am

jdamerow:
ok to test

Quadriga/src/main/webapp/WEB-INF/views/public/projectstatistics.jsp
```diff
@@ -61,6 +61,6 @@ document.body.onload = function()
 </div>
 <div class="col-md-6">
    <h3>Submitted Networks </h3>
-   <p>The diagram on the left shows how many networks have been submitted to this project in the past.
+   <p>The diagram on the left shows how many networks have been submitted to this project in the past one month.
```
jdamerow: should just be "past month"

jdamerow:
Probably not directly related to this story, but please also fix the following error (probably because there was no activity):

image

jdamerow:
ok to test

jdamerow:
almost. label for the current day is missing:
image

nischalsamji:
Think it is because of the total no. of dates and ticks. Looks fine today. Do you want me to add an extra tick that specifies current day?
screen shot 2016-09-29 at 11 55 24 am

jdamerow:
yes, every bar has to have  a label.

jdamerow:
Looking at it again, is there a reason only every other day has a label?

nischalsamji:
Looks like this after adding each day
screen shot 2016-10-03 at 9 04 31 am

jdamerow:
ok to test

PR #332, opened 2016-09-30
[QUAD-196]: Added the ability to view all the public networks in quadriga.

suprajkolluri:
Added the ability to view all the public networks in quadriga in a single page in quadriga/sites section.


Quadriga/src/main/java/edu/asu/spring/quadriga/dao/impl/NetworkDAO.java
```diff
@@ -85,7 +74,7 @@ public NetworksDTO getNetworksDTO(String networkId) throws QuadrigaStorageExcept
         try {
             networksDTO = (NetworksDTO) sessionFactory.getCurrentSession().get(NetworksDTO.class, networkId);
             return networksDTO;
-        } catch (Exception e) {
+        } catch (HibernateException e) {
             logger.error("getNetworksDTO error :", e);
```
jdamerow, line 78: remove logger calls. the exception should be logged when it's handled

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/open/NodeInfoController.java
```diff
@@ -26,26 +26,27 @@
...
+    public ResponseEntity<String> getTextsForConcepts(@RequestParam String conceptId, @RequestParam String projectUnix)
+            throws Exception {
         Set<TextOccurance> occurances = networkManager.getTextsForConceptId(conceptId, ETextAccessibility.PUBLIC);
         JSONArray projectTexts = new JSONArray();
         JSONArray otherProjectsTexts = new JSONArray();
-        
+
         for (TextOccurance occur : occurances) {
-            if (occur.getProject().getUnixName().equals(projectUnix.trim())) {
+            if ((occur.getProject().getUnixName().equals(projectUnix.trim()) || projectUnix.isEmpty())
```
jdamerow, line 44: you probably still want to call `trim` on the project unix name before calling `isEmpty`

Quadriga/src/main/webapp/WEB-INF/tiles/layouts/skeleton_public_sites.jsp
```diff
@@ -105,21 +110,39 @@
...
+			</div>
+		</div>
+
+		<div class="col-md-3" style="text-align: center">
+			<!-- Sidebar -->
+			<!-- Recent Posts -->
+			<tiles:insertAttribute name="right-navigation" />
+			<!-- /Recent Posts -->
+			<!-- /Sidebar -->
+		</div>
```
jdamerow, line 145: Is this div always on the page? Even if there is no right navigation? That's not good. Why don't you just use two separate skeletons?

Quadriga/src/main/java/edu/asu/spring/quadriga/web/sites/SitesController.java
```diff
@@ -11,53 +11,100 @@
...
+
+    /**
+     * This method loads all the public networks and navigates to the public
+     * networks page to display the graph
+     * 
+     * @param model
+     * @return
+     * @throws QuadrigaStorageException
+     */
+    @RequestMapping(value = "sites/network", method = RequestMethod.GET)
```
jdamerow, line 92: this method should have it's own controller

jdamerow (changes requested):

PR #335, opened 2016-10-04
Story/quad 136

nischalsamji:
Adding Functionality to add network images to blog posts


Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
```diff
@@ -132,15 +152,51 @@ public ModelAndView addProjectBlogEntryRequest(
...
+    public ResponseEntity<String> visualizeNetworks(@ProjectIdentifier @PathVariable("projectUnixName") String unixName,
+            @PathVariable("networkId") String networkId, Principal principal,
+            @CheckAccess @InjectProject IProject project) throws QuadrigaStorageException, JAXBException {
+
+        ITransformedNetwork transformedNetwork = transformationManager.getTransformedNetwork(networkId);
+        String json = null;
+        if (transformedNetwork != null) {
+            json = jsonCreator.getJson(transformedNetwork.getNodes(), transformedNetwork.getLinks());
+        }
+        return new ResponseEntity<String>(json, HttpStatus.OK);
```
jdamerow, line 215: if there is no json (because there was an error retrieving a network) you shouldn't return an OK status. Also, since this is a method that doesn't return a complete page but json but is not a method of a REST controller, you need to handle exception and return proper http statuses so that your javascript can handle it appropriately.

Quadriga/src/main/webapp/WEB-INF/tiles/layouts/skeleton_public.jsp
```diff
@@ -1,87 +1,120 @@
...
+<script
+	src="${pageContext.servletContext.contextPath}/resources/txt-layout/js/d3networkvisualize.js"></script>
+<script src="https://d3js.org/d3.v3.js" charset="utf-8"></script>
+<script
+	src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.3/jquery.min.js"></script>
+<script>
+	window.jQuery
+			|| document
+					.write('<script src="../../assets/js/vendor/jquery.min.js"><\/script>')
+</script>
```
jdamerow: indentation

Quadriga/src/main/webapp/WEB-INF/tiles/layouts/skeleton_public.jsp
```diff
@@ -1,87 +1,120 @@
...
+	src="${pageContext.servletContext.contextPath}/resources/js/bootstrap-3.3.6-dist/js/bootstrap.min.js"></script>
+</head>
+
+<body>
+	<tiles:importAttribute name="currentPage" scope="request" />
+	<tiles:importAttribute name="showNetwork" scope="request" />
+
+	<c:if test="${showNetwork != 'False'}">
+		<div class="image-bg">
+			<!-- <img src="${pageContext.servletContext.contextPath}/resources/public/imgs/nw-big-5.png"> -->
```
jdamerow: what's up with this?

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
```diff
@@ -35,33 +13,193 @@
...
+					ed.on('init', function() {
+						this.getDoc().body.style.fontSize = '14px';
+					});
+				}
+
+			});
+</script>
+<script src="/quadriga/resources/js/d3.min.js" charset="utf-8"
+	type="text/javascript"></script>
 <!-- /Content -->
```
jdamerow, line 234: fix indentation for this file

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
```diff
@@ -35,33 +13,193 @@
...
+			datatype : 'text',
+			url : "${pageContext.servletContext.contextPath}/sites/"
+					+ "sampleproj" + "/visualizenetwork/" + 'NETCEnz9o',
+			timeout : 100000,
+			success : function(data) {
+				visualizeNetwork(data);
+			},
+			error : function(e) {
+				console.log(e.Message);
+			}
```
jdamerow: the user should see an actual error message if a network couldn't be retrieve for some reason.

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
```diff
@@ -35,33 +13,193 @@
...
+	src="${pageContext.servletContext.contextPath}/resources/js/cytoscape/dist/cytoscape.js"
+	type="text/javascript"></script>
+<script type="text/javascript">
+	function loadNetwork() {
+		$.ajax({
+			type : "GET",
+			contentType : "application/json",
+			datatype : 'text',
+			url : "${pageContext.servletContext.contextPath}/sites/"
+					+ "sampleproj" + "/visualizenetwork/" + 'NETCEnz9o',
```
jdamerow: what's up with the "sampleproj"?

jdamerow: and network id?

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
```diff
@@ -35,33 +13,193 @@
...
+											'click',
+											function() {
+												var png = cy.png({
+													'scale' : 0.75
+												});
+												tinyMCE
+														.execCommand(
+																'mceInsertContent',
+																false,
+																'<img alt="Smiley face" src="' + png + '"/>');
```
jdamerow: why is the alt text for the image "Smiley face"?

Quadriga/src/main/webapp/resources/public/css/base.css
```diff
@@ -52,4 +52,12 @@
...
 span.highlight {
     background-color: rgb(206, 167, 24);
+}
+
+
+#networkView{
+		width: 100%;
+        height: 30%;
+        top: 0px;
+        left: 0px;
```
jdamerow: indentation

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
```diff
@@ -132,15 +152,53 @@ public ModelAndView addProjectBlogEntryRequest(
...
+            @PathVariable("networkId") String networkId, Principal principal, @InjectProject IProject project,
+            @RequestParam("projectId") String projectId) throws QuadrigaStorageException {
+
+        ITransformedNetwork transformedNetwork = transformationManager.getTransformedNetwork(networkId);
+        String json = null;
+        if (transformedNetwork != null) {
+            json = jsonCreator.getJson(transformedNetwork.getNodes(), transformedNetwork.getLinks());
+        } else
+            return new ResponseEntity<String>(json, HttpStatus.NO_CONTENT);
+
```
jdamerow, line 214: missing brackets

jdamerow: line 200, return should be not found.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
```diff
@@ -132,15 +152,53 @@ public ModelAndView addProjectBlogEntryRequest(
...
+     *            principal object which is required to fetch information about
+     *            logged in user.
+     * @param project
+     *            project instance obtained using @InjectProject annotation
+     * @return Returns a JSON string as a response entity based on the network
+     *         selected from the UI.
+     * @throws QuadrigaStorageException
+     * @throws JAXBException
+     */
+    @AccessPolicies({ @ElementAccessPolicy(type = CheckedElementType.PROJECT, paramIndex = 5, userRole = {
```
jdamerow, line 193: the access policy is for sites that require user authentication. pages under 'sites' are open (unless the project is not public). You probably want to use @InjectProjectByName, but make sure it checks that the project is public.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
```diff
@@ -132,15 +152,53 @@ public ModelAndView addProjectBlogEntryRequest(
...
+            RoleNames.ROLE_COLLABORATOR_OWNER, RoleNames.ROLE_PROJ_COLLABORATOR_ADMIN }) })
+    @InjectProjectByName
+    @RequestMapping(value = "sites/{projectUnixName}/visualizenetwork/{networkId}", method = RequestMethod.GET, produces = "text/plain")
+    public ResponseEntity<String> visualizeNetworks(@ProjectIdentifier @PathVariable("projectUnixName") String unixName,
+            @PathVariable("networkId") String networkId, Principal principal, @InjectProject IProject project,
+            @RequestParam("projectId") String projectId) throws QuadrigaStorageException {
```
jdamerow: QuadrigaStorageException is thrown by the transformation manager, right? You need to handle the exception here.

Quadriga/src/main/webapp/WEB-INF/tiles/layouts/skeleton_public.jsp
jdamerow: please remove the comment, I think I remember why it's there. just remove it.

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
jdamerow: please fix indentation

Quadriga/src/main/webapp/resources/public/css/base.css
```diff
@@ -21,35 +21,40 @@
...
-    background-color: rgb(206, 167, 24);
+	background-color: rgb(206, 167, 24);
+}
+
+#networkView {
+	width: 100%;
+	height: 30%;
+	top: 0px;
+	left: 0px;
 }
```
jdamerow, line 61: please keep previous indentation

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
nischalsamji: I was confused here, because only authenticated users can add a blog post, I used these access checks, is it necessary to check if the project is public too?

jdamerow: hmm, thinking about that again, this method is only used for creating a block entry, right? In that case this should work. If it's used to display networks anywhere else, then the authorization as it is done right now would be a problem. Let's add to the javadoc that this method works only for authorized users. We need to clean the network visualization up eventually, but for now this will probably work.

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
jdamerow: still not fixed

jdamerow (changes requested):

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
nischalsamji: I am using the default formatter for jsps after replacing tabs with spaces. Is there anything that I need to change in the formatter?

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
```diff
@@ -132,15 +157,60 @@ public ModelAndView addProjectBlogEntryRequest(
...
+    @RequestMapping(value = "sites/{projectUnixName}/visualizenetwork/{networkId}", method = RequestMethod.GET, produces = "text/plain")
+    public ResponseEntity<String> visualizeNetworks(@ProjectIdentifier @PathVariable("projectUnixName") String unixName,
+            @PathVariable("networkId") String networkId, Principal principal,
+            @CheckAccess @InjectProject IProject project, @RequestParam("projectId") String projectId) {
+        ITransformedNetwork transformedNetwork = null;
+        try {
+            transformedNetwork = transformationManager.getTransformedNetwork(networkId);
+        } catch (QuadrigaStorageException qse) {
+
+            logger.error("Error:" + qse.getMessage());
```
jdamerow: please log the whole stack trace. Otherwise, this won't be very helpful

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
```diff
@@ -35,33 +13,211 @@
...
+
+
+						</div>
+					</div>
+				</c:when>
+				<c:when test="${empty networks}">
+					<p>There are no networks in this project.</p>
+				</c:when>
+			</c:choose>
+		</div>
```
jdamerow: this is hardly readable with this indentation width. please reduce indentation.

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
```diff
@@ -35,33 +13,211 @@
...
+	type="text/javascript"></script>
+<script
+	src="${pageContext.servletContext.contextPath}/resources/js/cytoscape/publicNetwork.js"
+	type="text/javascript"></script>
+<script
+	src="${pageContext.servletContext.contextPath}/resources/js/cytoscape/dist/cytoscape.js"
+	type="text/javascript"></script>
+<script type="text/javascript">
+    function loadNetwork(selectedNW) {
+        var projid = $('#projectId').val();
```
jdamerow: why don't you use `${project.projectId}` here (and move it down where it's needed)? It should only show networks of this project.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
```diff
@@ -132,15 +157,60 @@ public ModelAndView addProjectBlogEntryRequest(
...
+     * @throws QuadrigaStorageException
+     * @throws JAXBException
+     */
+    @AccessPolicies({ @ElementAccessPolicy(type = CheckedElementType.PROJECT, paramIndex = 5, userRole = {
+            RoleNames.ROLE_COLLABORATOR_OWNER, RoleNames.ROLE_PROJ_COLLABORATOR_ADMIN }) })
+    @InjectProjectByName
+    @RequestMapping(value = "sites/{projectUnixName}/visualizenetwork/{networkId}", method = RequestMethod.GET, produces = "text/plain")
+    public ResponseEntity<String> visualizeNetworks(@ProjectIdentifier @PathVariable("projectUnixName") String unixName,
+            @PathVariable("networkId") String networkId, Principal principal,
+            @CheckAccess @InjectProject IProject project, @RequestParam("projectId") String projectId) {
```
jdamerow, line 199: why do you need the projectId parameter here? You can get the id from the injected project.

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
```diff
@@ -35,33 +13,211 @@
...
+        var projName = $('#projectName').val();
+        var nwid = selectedNW.getAttribute('value');
+        $.ajax({
+            type : "GET",
+            data : {
+                projectId : projid
```
jdamerow: see my comment further up, I don't think you need this.

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
```diff
@@ -35,33 +13,211 @@
...
+                loadErrorMessage();
+            }
+        });
+
+    }
+
+    function loadErrorMessage() {
+        $('#networkBox')
+                .append(
+                        "<p>There was an error while loading this network. Please contact an administrator</p>");
```
jdamerow: missing period at the end

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
jdamerow: ?

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
```diff
@@ -1,67 +1,234 @@
...
+<script
+    src="${pageContext.servletContext.contextPath}/resources/js/cytoscape/dist/cytoscape.js"
+    type="text/javascript"></script>
+<script type="text/javascript">
+    function loadNetwork(selectedNW) {
+        var nwid = selectedNW.getAttribute('value');
+        $.ajax({
+            type : "GET",
+            data : {
+                projectId : ${project.projectId}
```
jdamerow: why is that here? The method doesn't take any request parameters anymore, does it?

jdamerow:
Nischal, you really need to pay more attention to the comments and react to all of them (if something can't be change, or you think shouldn't, you have to respond to the comment at least.)

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
nischalsamji: I used this for access checks. Is there a way I can check access without using the projectid as a parameter?

Quadriga/src/main/webapp/WEB-INF/views/addprojectblogentry.jsp
nischalsamji: I used this for access checks at the controller level.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
jdamerow: You will have to write a new aspect to check that. Just write one that works similar to the AccessPolicies but takes an object rather than the id and checks the object directly. You can then use that for all methods in this class. The way it is right now, it's not secure. I could just call the url for one project but with the id of a different project, and this way could have access to a project that I shouldn't have. Does this make sense?

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
```diff
@@ -45,11 +61,22 @@
...
     @Autowired
     private IProjectBlogEntryManager projectBlogEntryManager;
 
+    @Autowired
+    private INetworkTransformationManager transformationManager;
+
+    @Autowired
+    private IJsonCreator jsonCreator;
+
+    private static final Logger logger = LoggerFactory.getLogger(AddProjectBlogEntryController.class);
```
jdamerow, line 78: that should not be static

Quadriga/src/main/java/edu/asu/spring/quadriga/web/projectblog/AddProjectBlogEntryController.java
```diff
@@ -132,15 +157,61 @@ public ModelAndView addProjectBlogEntryRequest(
...
+    @RequestMapping(value = "sites/{projectUnixName}/visualizenetwork/{networkId}", method = RequestMethod.GET, produces = "text/plain")
+    public ResponseEntity<String> visualizeNetworks(@ProjectIdentifier @PathVariable("projectUnixName") String unixName,
+            @PathVariable("networkId") String networkId, Principal principal,
+            @CheckAccess @InjectProject IProject project, @RequestParam("projectId") String projectId) {
+        ITransformedNetwork transformedNetwork = null;
+        try {
+            transformedNetwork = transformationManager.getTransformedNetwork(networkId);
+        } catch (QuadrigaStorageException qse) {
+
+            logger.error("Error:" + qse);
```
jdamerow, line 205: this is not how you log errors

PR #339, opened 2016-10-18
[QUAD-231]: Fix Full screen issue when screen gets resized.

nischalsamji:
Fixed button issue when screen is resized.


jdamerow:
ok to test

PR #341, opened 2016-10-22
[QUAD-234] Fixing the bug

suprajkolluri:


Quadriga/src/main/java/edu/asu/spring/quadriga/web/dictionary/DictionaryItemSearchAddController.java
```diff
@@ -96,29 +90,64 @@ public String addDictionaryItem(HttpServletRequest req, @PathVariable("dictionar
...
+        if (owner) {
+            model.addAttribute("isAdmin", true);
+            model.addAttribute("hasWrite", true);
+            model.addAttribute("hasRead", true);
+        } else {
+            List<IDictionaryCollaborator> existingCollaborators = dictionaryManager
+                    .showCollaboratingUsers(dictionaryId);
+            setPermissions(model, userName, existingCollaborators);
+        }
+
```
jdamerow: why don't you just redirect to the dictionary page with a success message instead of duplicating the code?

jdamerow:
ok to test

PR #342, opened 2016-10-24
Story/quad 136

nischalsamji:


jdamerow:
Please clean up your code (it seems to be copy-pasted and has wrong javadoc, variable names, formatting, etc. etc.) and resolve conflicts before I review it.

PR #344, opened 2016-10-24
Fix invalid tags in PublicWebsiteEditAbout PAge

nischalsamji:
Created a new branch for the bug in sync with Jira.


jdamerow:
haha, well you could have kept the previous one (my comment was for next time). but this works too ;)

jdamerow:
ok to test

jdamerow:
Since you're at it. Would you please fix the submit button as well? (Clicking save returns a 405 error).

PR #345, opened 2016-10-26
[QUAD-236] : Validate input XML and return correct error message

snilapwa:


Quadriga/src/main/java/edu/asu/spring/quadriga/service/passthroughproject/impl/XMLReader.java
```diff
@@ -101,7 +104,7 @@ public XMLInfo getXMLInfo(String xml) throws DocumentParserException {
         try {
             document = getXMLDocument(xml);
         } catch (ParserConfigurationException | IOException | SAXException e) {
-            throw new DocumentParserException(e);
+            throw new DocumentParserException(e.getMessage());
```
jdamerow: don't swallow the stacktrace

Quadriga/src/main/java/edu/asu/spring/quadriga/service/passthroughproject/impl/XMLReader.java
```diff
@@ -68,14 +68,17 @@ public String getNetwork(String xml) {
...
+    private String getTagId(Document document, String tagName) throws DocumentParserException {
 
         Node node = document.getElementsByTagName(tagName).item(0);
         if (node == null) {
             return null;
         }
         NamedNodeMap nodeAttributeMap = node.getAttributes();
         Node idNode = nodeAttributeMap.getNamedItem("id");
+        if(idNode == null) {
+            throw new DocumentParserException("Please provide a " + tagName + " id as a part of the XML");
```
jdamerow: Should be "Please provide an id attribute for tag: tagname".

jdamerow: And actually, I don't think this is correct. Please check the logic, but I think a user can create a project or workspace by submitting no ids.

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/service/passthroughproject/impl/XMLReader.java
snilapwa: Yes, User can upload without providing project id and/or workspace id. In that case user should not include project/workspace tag in input XML. But, if user add tag 'project' and/or 'workspace' and do not provide id for the tag, we get null pointer exception.

jdamerow (approved):

jdamerow:
ok to test

Quadriga/src/main/java/edu/asu/spring/quadriga/service/passthroughproject/impl/XMLReader.java
```diff
@@ -68,14 +68,17 @@ public String getNetwork(String xml) {
...
 
         Node node = document.getElementsByTagName(tagName).item(0);
         if (node == null) {
             return null;
         }
         NamedNodeMap nodeAttributeMap = node.getAttributes();
         Node idNode = nodeAttributeMap.getNamedItem("id");
+        if(idNode == null) {
+            throw new DocumentParserException("Please provide an id attribute for tag: " + tagName);
+        }
```
jdamerow, line 81: Generally using exceptions to validate input is not good practice. But for now let's leave it like that.

PR #346, opened 2016-10-28
Story/quad 227

snilapwa:
Test after CCP-163 gets merged to develop branch

### DIFF
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/ConceptpowerReply.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/ConceptpowerReply.java
@@ -48,7 +48,6 @@ import javax.xml.bind.annotation.XmlType;
  *         &lt;/element>
  *         &lt;element name="alternativeIds" maxOccurs="unbounded">
  *           &lt;element name="id" type="{http://www.w3.org/2001/XMLSchema}string"/>
- *           &lt;element name="id" type="{http://www.w3.org/2001/XMLSchema}string"/>
  *         &lt;/element>
  *       &lt;/sequence>
  *     &lt;/restriction>
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/ConceptpowerReply.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/ConceptpowerReply.java
@@ -41,13 +41,13 @@ import javax.xml.bind.annotation.XmlType;
  *                   &lt;element name="description" type="{http://www.w3.org/2001/XMLSchema}string"/>
  *                   &lt;element name="conceptList" type="{http://www.w3.org/2001/XMLSchema}string"/>
  *                   &lt;element name="type" type="{http://www.w3.org/2001/XMLSchema}anyType"/>
+ *                   &lt;element name="alternativeIds" maxOccurs="unbounded">
+ *                     &lt;element name="id" type="{http://www.w3.org/2001/XMLSchema}string"/>
+ *                   &lt;/element>
  *                 &lt;/sequence>
  *               &lt;/restriction>
  *             &lt;/complexContent>
  *           &lt;/complexType>
- *         &lt;/element>
- *         &lt;element name="alternativeIds" maxOccurs="unbounded">
- *           &lt;element name="id" type="{http://www.w3.org/2001/XMLSchema}string"/>
  *         &lt;/element>
  *       &lt;/sequence>
  *     &lt;/restriction>
@@ -59,7 +59,7 @@ import javax.xml.bind.annotation.XmlType;
  */
 @XmlAccessorType(XmlAccessType.FIELD)
 @XmlType(name = "", propOrder = {
-    "conceptEntry", "alternativeIds"
+    "conceptEntry"
 })
 @XmlRootElement(name = "conceptpowerReply")
 public class ConceptpowerReply {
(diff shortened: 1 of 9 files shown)


Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/ConceptpowerReply.java
```diff
@@ -96,7 +99,29 @@
     public void setConceptEntry(List<ConceptpowerReply.ConceptEntry> entries) {
         this.conceptEntry = entries;
     }
+    
+    @XmlElement(required = true, namespace="http://www.digitalhps.org/")
+    protected List<ConceptpowerReply.AlternativeIds> alternativeIds;
```
jdamerow: This does not have to be a list, does it? There should only be one element alternativeIds, right?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/transform/impl/EventGraphMapper.java
```diff
@@ -291,6 +305,10 @@ protected Node copyGraph(Node start) {
 			((Relation) copiedNode).setPredicate(copyGraph(predicate));
 			((Relation) copiedNode).setObject(copyGraph(object));
 		}
+		else {
+		    copiedNode.setAlternativeIds(start.getAlternativeIds());
```
jdamerow: Since alternativeIds is a property of node, you can just put that in line 198. no need for an else.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/transform/impl/Node.java
```diff
@@ -56,6 +57,12 @@ public String getEventId() {
 	public void setEventId(String eventId) {
 		this.eventId = eventId;
 	}
+    public List<String> getAlternativeIds() {
+        return alternativeIds;
+    }
+    public void setAlternativeIds(List<String> alternativeIds) {
+        this.alternativeIds = alternativeIds;
+    }
```
jdamerow, line 74: formatting is off. if the file still uses 6 spaces, just format the whole file

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/transform/impl/PatternFinder.java
```diff
@@ -78,8 +78,21 @@ public boolean equals(Node graph, Node pattern) {
 					return true;
 				}
 				if (graph.getConcept()!=null && pattern.getConcept() != null) {
-					if (!graph.getConcept().equals(pattern.getConcept()))
-						return false;
+					boolean isFound = false;
+					if (!graph.getConcept().equals(pattern.getConcept())) {
+					    // check for alternate ids
+					    if(graph.getAlternativeIds() != null) {
```
jdamerow: Doesn't alternative ids contain all ids (including the main one)? If so, you can remove line 82 and just go through all alternative ids.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/transform/impl/PatternFinder.java
```diff
@@ -78,8 +78,21 @@ public boolean equals(Node graph, Node pattern) {
...
+					        for(String alternateId : graph.getAlternativeIds()) {
+					            if(alternateId.equals(pattern.getConcept())) {
+					                isFound = true;
+					                break;
+					            }
+					        }
+					    }
```
jdamerow: why don't you use `contains` on the alternative id list?

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/ConceptpowerReply.java
snilapwa: There is one alternativeIds entry for each conceptEntry in ConceptpowerReply. As conceptEntry is defined as List, alternativeIds is defined as list.

snilapwa:
Corrected code as per review comments.

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/ConceptpowerReply.java
jdamerow: That doesn't make sense. AlternativeIds is a property of ConceptEntry. Then the class structure is wrong. Otherwise you would endup with a list of concept entries and a list of alternative ids, but you want each entry to have a list of alternative ids. You probably don't even need its own class. Just parse the elements inside alternativeIds into a list of strings (holding the ids).

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/transform/impl/PatternFinder.java
```diff
@@ -78,23 +78,13 @@ public boolean equals(Node graph, Node pattern) {
...
-					        for(String alternateId : graph.getAlternativeIds()) {
-					            if(alternateId.equals(pattern.getConcept())) {
-					                isFound = true;
-					                break;
-					            }
-					        }
-					    }
-					    if(!isFound) {
-					        return false;
+					if(graph.getAlternativeIds() != null) {
```
jdamerow: This should be in your if in line 80. Why two if statements?

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/impl/ConceptpowerReply.java
```diff
@@ -200,6 +181,9 @@ public void setWordnetId(String wordnetId) {
         @XmlElement(namespace="http://www.digitalhps.org/", type=ConceptpowerType.class)
         protected ConceptpowerType type;
 
+        @XmlElement(required = true, namespace="http://www.digitalhps.org/")
+        protected AlternativeIds alternativeIds;
```
jdamerow: have you tried using an `@XmlElementWrapper` annotation here and avoid having an extra class just for a list of strings? Just put it right into a list of strings?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/transform/impl/PatternFinder.java
```diff
@@ -77,14 +77,12 @@ public boolean equals(Node graph, Node pattern) {
...
 				}
-				if (graph.getConcept()!=null && pattern.getConcept() != null) {
-					if(graph.getAlternativeIds() != null) {
-					    if(graph.getAlternativeIds().contains(pattern.getConcept())) {
-					        graph.setId(pattern.getId());
-					        return true;
-					    }
-					}
-					return false;
+				if (graph.getConcept() != null && pattern.getConcept() != null && graph.getAlternativeIds() != null) {
```
jdamerow: since you're not using `graph.getConcept` in that if block, why do you check if it's null?

jdamerow (changes requested):

jdamerow:
ok to test

jdamerow (approved):

PR #348, opened 2016-11-03
[QUAD-232]:Update deactivate workspace and delete project to new layout

snilapwa:


jdamerow:
ok to test

jdamerow:
The list of workspaces should have the new layout as well (as on the main workspace page). Should not look like this:
image

jdamerow:
ok to test

PR #353, opened 2016-12-03
[QUAD-242] On quadriga/sites/network (public sites), filtering the data

suprajkolluri:
On quadriga/sites/network (public sites), filtering the data to retrieve networks of only public projects


jdamerow:
ok to test

PR #355, opened 2017-02-09
Story/quad 229

suprajkolluri:
Gets a high level network of the most repeated terms.


Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkManager.java
```diff
@@ -455,6 +466,28 @@ public String storeNetworkDetails(String xml, IUser user, String networkName, St
     }
 
     /**
+     * {@inheritDoc}
+     */
+    @Override
+    public List<CreationEvent> getTopElementEvents(String xml) throws JAXBException {
+        ElementEventsType elementEventType = marshallingService.unMarshalXmlToElementEventsType(xml);
+        NewNetworkDetailsCache newNetworkDetailCache = new NewNetworkDetailsCache();
```
jdamerow: give that you replace the new instance in the next line, the instantiation is unnecessary.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkManager.java
```diff
@@ -455,6 +466,28 @@ public String storeNetworkDetails(String xml, IUser user, String networkName, St
...
+
+        newNetworkDetailCache = parseNewNetworkStatement(elementEventType, newNetworkDetailCache);
+
+        List<NetworkEntry> networkDetailsCache = newNetworkDetailCache.getEntries();
+        List<String> topIDs = networkDetailsCache.stream().filter(entry -> entry.isTop()).map(entry -> entry.getId())
+                .collect(Collectors.toList());
+
+        List<CreationEvent> eventList = elementEventType.getRelationEventOrAppellationEvent();
+
+        return eventList.stream().filter(event -> topIDs.contains(event.getId())).collect(Collectors.toList());
```
jdamerow: are you filtering out non-public projects?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkManager.java
```diff
@@ -284,20 +287,28 @@ public INetwork getNetwork(String networkId) throws QuadrigaStorageException {
...
     public List<INetwork> getApprovedNetworkList() throws QuadrigaStorageException {
         return networkmapper.getListOfApprovedNetworks();
     }
 
     /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getNetworkWithPopularTerms() throws QStoreStorageException {
+        return qStoreConnector.executeQuery(env.getProperty("allNetworks"), RELATION_EVENT);
```
jdamerow: I would put the query part on the qstoreconnector level. This class should be as oblivious as possible regarding specific storage imlementation.

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
```diff
@@ -174,16 +204,44 @@ public String store(String xml) throws QStoreStorageException {
...
+            throw new QStoreStorageException(e);
+        }
+        return res;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String executeQuery() throws QStoreStorageException {
```
jdamerow: this method sounds like it would be provided with a query and then execute it (this is actually what you say in the javadoc), but there is a specific query that you load from the properties file and execute.

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
```diff
@@ -174,16 +204,44 @@ public String store(String xml) throws QStoreStorageException {
...
+        String query = env.getProperty("allNetworks");
+        String res = "";
```
jdamerow: instantiate/initialize when needed

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
```diff
@@ -174,16 +204,44 @@ public String store(String xml) throws QStoreStorageException {
...
+        RestTemplate restTemplate = new RestTemplate();
+
+        HttpHeaders headers = buildRestHeader(messageConverters, restTemplate);
+        HttpEntity<String> request = new HttpEntity<String>(query, headers);
+
+        try {
+            // execute the query in Qstore and get the result
+            String url = getQStoreQueryURL();
+            UriComponentsBuilder builder = UriComponentsBuilder.fromHttpUrl(url).queryParam("class", RELATION_EVENT);
+            res = restTemplate.postForObject(builder.build().encode().toUri(), request, String.class);
```
jdamerow: return here and you don't even need a ```res``` variable.

jdamerow (changes requested):

jdamerow (approved):

jdamerow:
ok to test

jdamerow:
This seems to take so long that Apache times out.

suprajkolluri:
@jdamerow Can you check now, I replaced List with a Set in NetworkManager.java line 470.
During filtering this will reduce the complexity for 'topIDs.contains(id)' to O(1) from O(n).

jdamerow:
still proxy error, but looks like this time it's timing out?

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
```diff
@@ -74,91 +80,129 @@
     private String qStoreURL_Get;
 
     @Autowired
+    @Qualifier("qStoreURL_Query")
+    private String qStoreURL_Query;
+
+    @Autowired
+    @Qualifier("qStoreURL_Aysnc_Query_Result")
+    private String qStoreURL_Aysnc_Query_Result;
```
jdamerow: Let's name those variables according to the coding style we use (I know the ones that are already there haven't been following that either, but they should).

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
```diff
@@ -174,16 +218,86 @@ public String store(String xml) throws QStoreStorageException {
...
+        String query = env.getProperty("allNetworks");
+        String queryID = env.getProperty("allNetworks_id");
+
+        executeQuery(query, queryID);
+
+        long delay = Long.parseLong(env.getProperty("qstore.rest.delay"));
+        String res = getQueryResult(queryID);
+
+        // Keep polling QStore until we get the result
+        while (res != null && res.contains("<message>") && res.contains(RUNNING.name())) {
```
jdamerow: you should parse the response and use the parse values for comparison

Quadriga/src/main/java/edu/asu/spring/quadriga/web/sites/PublicNetworkController.java
```diff
@@ -1,56 +1,97 @@
...
+
+    /**
+     * This method returns a JSON String that represents all the public networks
+     * 
+     * @throws QStoreStorageException
+     * @throws JAXBException
+     * @throws QuadrigaStorageException
+     * 
+     */
+    @RequestMapping(value = "public/rest/networks", method = RequestMethod.GET)
```
jdamerow: I wouldn't call these 'rest'  as they are not part of the REST api

Quadriga/src/main/java/edu/asu/spring/quadriga/web/sites/PublicNetworkController.java
```diff
@@ -1,56 +1,97 @@
...
+    /**
+     * This method tells if the asynchronous query has loaded the public
+     * networks or not
+     * 
+     * @return
+     */
+    @RequestMapping(value = "public/rest/networkloadstatus", method = RequestMethod.GET)
+    public ResponseEntity<String> getPublicNetworkLoadStatus() {
+        if (qStoreAsyncTaskManager.getPublicNetworkStatus()) {
+            return new ResponseEntity<>(COMPLETED.name(), HttpStatus.OK);
```
jdamerow: what if there was an error in QStore? (but no exception, just an error message coming back from qstore)

Quadriga/src/main/resources/ehcache.xml
```diff
@@ -23,5 +23,9 @@
     <cache name="conceptSearch" eternal="false" maxElementsInMemory="2000"
         overflowToDisk="false" diskPersistent="false" timeToIdleSeconds="0"
         timeToLiveSeconds="21600" memoryStoreEvictionPolicy="LRU" />
+        
+	<cache name="publicNetworksStatus" eternal="false" maxElementsInMemory="1000"
```
jdamerow: indentation

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/async/impl/QStoreAsyncTaskManager.java
```diff
@@ -0,0 +1,56 @@
...
+    private Future<String> publicNetworks;
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    @Cacheable(value = "publicNetworksStatus")
+    public String startLoadingPublicNetworks() throws QStoreStorageException {
+        publicNetworks = qStoreConnector.loadNetworkWithPopularTerms();
+        return RUNNING.name();
```
jdamerow: why is this not returning the Enum?

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/async/impl/QStoreAsyncTaskManager.java
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String getNetworkWithPopularTerms() throws QStoreStorageException {
+        try {
+            return publicNetworks.get();
+        } catch (InterruptedException | ExecutionException e) {
+            throw new QStoreStorageException(e);
```
jdamerow: why would that throw a storage exception? It's not trying to store anythign.

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
```diff
@@ -174,16 +215,104 @@ public String store(String xml) throws QStoreStorageException {
...
+        HttpHeaders headers = buildRestHeader(messageConverters, restTemplate);
+        HttpEntity<String> request = new HttpEntity<String>(query, headers);
+
+        try {
+            // execute the query in Qstore and get the result
+            String url = getQStoreQueryURL();
+            UriComponentsBuilder builder = UriComponentsBuilder.fromHttpUrl(url).queryParam("class", RELATION_EVENT);
+            return restTemplate.postForEntity(builder.build().encode().toUri(), request, String.class);
+        } catch (RestClientException e) {
+            throw new QStoreStorageException(e);
```
jdamerow: same here: why storage exception?

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
```diff
@@ -174,16 +215,104 @@ public String store(String xml) throws QStoreStorageException {
...
+
+        HttpHeaders headers = buildRestHeader(messageConverters, restTemplate);
+        HttpEntity<String> request = new HttpEntity<String>(headers);
+
+        try {
+            String url = qStoreURL + pollURL;
+            return restTemplate.exchange(url, HttpMethod.GET, request, String.class).getBody();
+
+        } catch (RestClientException e) {
+            throw new QStoreStorageException(e);
```
jdamerow: same here

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
```diff
@@ -284,12 +413,9 @@ protected HttpHeaders buildRestHeader(List<HttpMessageConverter<?>> messageConve
...
     }
 
-
     private String getAuthHeader() {
-        String auth = env.getProperty("qstore.admin.username") + ":"
-                + env.getProperty("qstore.admin.password");
-        byte[] encodedAuth = Base64.encodeBase64(auth.getBytes(Charset
-                .forName("US-ASCII")));
+        String auth = env.getProperty("qstore.admin.username") + ":" + env.getProperty("qstore.admin.password");
+        byte[] encodedAuth = Base64.encodeBase64(auth.getBytes(Charset.forName("US-ASCII")));
```
jdamerow: this is a bad idea, you can't necessarily make the assumption that this will be the correct charset. For now, I would just use the default platform charset, and we'll set only if this fails. (and then it should probably be utf-8)

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/async/QueryResultHandler.java
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    @Override
+    public Source marshal(String res, ValidationEventHandler errorHandler) {
+        String xml = PARAMETERS_START_TAG
+                + res.trim().replaceAll("&lt;", "<").replaceAll("&gt;", ">").replaceAll("&amp;", "&")
+                + PARAMETERS_END_TAG;
+        StringReader xmlReader = new StringReader(xml);
+        return new StreamSource(xmlReader);
+    }
```
jdamerow: see my comments on QStore PR

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/async/QueryResultHandler.java
suprajkolluri: We are the clients here right, we will always request an XML here, won't we?

jdamerow: right now we're the only clients. but even if we stay the only ones, we might want to switch to JSON eventually (which would be nice).

suprajkolluri: Ok, I'll make the changes.

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/async/impl/QStoreAsyncTaskManager.java
jdamerow: ?

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
```diff
@@ -174,16 +221,120 @@ public String store(String xml) throws QStoreStorageException {
...
+
+            StringReader reader = new StringReader(res);
+            return (QStoreAsyncResult) unmarshaller.unmarshal(reader);
+        } catch (JAXBException ex) {
+            throw new AsyncExecutionException("Invalid response XML from QStore " + res, ex);
+        }
+
+    }
+
+    private QStoreAsyncResult getResponseFromJSON(String res) throws AsyncExecutionException {
```
jdamerow: is that used anywhere?

jdamerow (changes requested):

jdamerow:
ok to test

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
suprajkolluri: Not now..but if we want to parse json result we can use this method.

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/async/QueryResultHandler.java
```diff
@@ -0,0 +1,57 @@
...
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+
+/**
+ * 
+ * Handler class to retrieve the query result from the QStore XML. Also converts
+ * character data in the result to markup characters.
+ *
+ */
+public class QueryResultHandler extends JsonDeserializer<String> implements DomHandler<String, StreamResult> {
```
jdamerow: this should be two different classes (one class = one responsibility)

Quadriga/src/main/java/edu/asu/spring/quadriga/qstore/impl/QStoreConnector.java
jdamerow: It's just a few lines of code, we can write that when we need it.

jdamerow:
let's show a loading spinner while the networks is loading
image

jdamerow (approved):

PR #356, opened 2017-02-16
[QUAD-244] Fixing the null string issue

suprajkolluri:
Fixing the null string issue while trying to cast the null string to an Integer.


jdamerow:
ok to test

PR #364, opened 2017-03-21
[QUAD-228] added fix to consider alternative ids of concept along with main concept id, for generating a network based on concept search

chiraag-subramanian:


Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -198,17 +196,34 @@ public ITransformedNetwork getTransformedNetworkusingNetworkList(List<INetwork>
...
         // add all the statement ids to a set
         Set<String> statementIdSearchSet = new HashSet<String>();
         List<Node> searchedNodes = new ArrayList<Node>();
         for (Node node : transformedNetwork.getNodes().values()) {
+            // check if the node's concept id matches the searched concept id
             if (conceptId.equals(node.getConceptId())) {
                 searchedNodes.add(node);
                 statementIdSearchSet.addAll(node.getStatementIds());
             }
+            // check if the node's concept id matches the alternative id of the concept
```
jdamerow: alternative ids should contain all ids (including the one that was searched for), so you can remove the previous if block.

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/SearchNetworkRestController.java
```diff
@@ -108,10 +113,19 @@
...
 
+        // Fetch ConceptPower entries related to the conceptId
+        ConceptpowerReply reply = conceptpowerConnector.getById(conceptId);
+        List<String> alternativeIdsForConcept = null;
+        if (reply != null && reply.getConceptEntry().size() > 0) {
+            alternativeIdsForConcept = reply.getConceptEntry().get(0).getAlternativeIdList();
+        }
+        if(alternativeIdsForConcept == null){
+            alternativeIdsForConcept = new ArrayList<String>();
+        }
```
jdamerow: there isn't really a reason to have this code here in the controller, instead of in the class that is actually using the alternative ids. Just creates duplicate code (with the other controller).

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
```diff
@@ -128,15 +128,39 @@ public String getSearch(@ProjectIdentifier @PathVariable("projectUnixName") Stri
...
+        String searchNodeLabel = "";
+        
+        // Fetch ConceptPower entries related to the conceptId
+        ConceptpowerReply reply = conceptpowerConnector.getById(conceptId);
+        List<String> alternativeIdsForConcept = null;
+        if (reply != null && reply.getConceptEntry().size() > 0) {
+            searchNodeLabel = reply.getConceptEntry().get(0).getLemma();
+            lemma = reply.getConceptEntry().get(0).getDescription();
+            alternativeIdsForConcept = reply.getConceptEntry().get(0).getAlternativeIdList();
+        }
```
jdamerow: concepts are cached, so it's not a big deal if the conceptconnector is called twice.

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/SearchNetworkRestController.java
chiraag-subramanian: Both the controllers SearchNetworkRestController.java and NetworkSearchController.java use the  service NetworkTransformationManager. Agree there is code duplication wrt finding alternative ids for a concept id. I thought fetching and providing information to the service layer is the work of the controller and the service layer is responsible for transformations and dealing with database layer. Wouldn't shifting the functionality (fetching alternative ids) to service layer lead to additional overhead in the service layer?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
chiraag-subramanian: Agree

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/SearchNetworkRestController.java
jdamerow: service layer is responsible for business logic, and that the alternative ids need to be used rather than should the main id is business logic. If you require every caller of the method to get the required ids from conceptpower, you a) duplicate code and b) put business logic into layers that shouldn't contain any. If the logic for some reason changes later, e.g. we'll have an "alternative ids" field and a "representing ids" field and we'll need to use "representing ids" rather than "alternative ids", then you would have to change every single caller of the business logic method to retrieve the correct ids, rather than once in the business logic. Make sense?

chiraag-subramanian: If it means abstracting away the "business-logic" then it makes sense to write the logic in the service layer.  Also as per MVC, the controller is responsible only for directing requests and returning the response.
There is a trade-off between abstraction and dealing with dependencies. Bit confused about the design approach to take in general. Will shift the logic to the Service layer.

jdamerow: Image it this way: if we were to switch frameworks (which has happened before) and not use Spring anymore, ideally (if we ignore that we've annotated all of our classes) we would just rewrite the controllers a bit to fit with another framework (preparing the data for the view). This would be really easy if all the controllers do is just preparing data for display, receiving data needed for executing logic and prepare the data for passing it to the service layer.

Not sure what you mean by  "trade-off between abstraction and dealing with dependencies". In the case right now, there isn't really a trade off. There are fewer dependencies to worry about in the controller and no abstraction is lost.

chiraag-subramanian: Thanks Julia. The explanation was helpful. Now I see the benefits of writing that logic in the service layer :)

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -90,18 +95,32 @@ public ITransformedNetwork getSearchTransformedNetwork(String projectId, String
...
-        // create finalnetwork using conceptId
-        return getFinalTransformedNetwork(transformedNetwork, conceptId);
+        
+        // create final network using alternativeIdsForConcept
+        return getFinalTransformedNetwork(transformedNetwork, alternativeIdsForConcept);
 
     }
-
+    
+   // Fetch ConceptPower entries related to the conceptId and return alternative id for the concept
```
jdamerow: should be proper javadoc

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -198,17 +219,27 @@ public ITransformedNetwork getTransformedNetworkusingNetworkList(List<INetwork>
         // return new network with updated nodes and links
         return new TransformedNetwork(updatedNodes, links);
     }
-
-    private ITransformedNetwork getFinalTransformedNetwork(ITransformedNetwork transformedNetwork, String conceptId) {
-        // Filter the nodes with the concept id
-        // add all the statement ids to a set
+    /**
+     * Filter the nodes in the network using the concept id and the alternative id of the concept
```
jdamerow: "alternative ids" (plural)

jdamerow: period missing at the end of the sentence

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -198,17 +219,27 @@ public ITransformedNetwork getTransformedNetworkusingNetworkList(List<INetwork>
...
+        // Add all the statement id of the selected node to a set
+        // The alternative id list if not empty, is sure to contain the concept id corresponding to the searched concept
         Set<String> statementIdSearchSet = new HashSet<String>();
         List<Node> searchedNodes = new ArrayList<Node>();
         for (Node node : transformedNetwork.getNodes().values()) {
-            if (conceptId.equals(node.getConceptId())) {
-                searchedNodes.add(node);
-                statementIdSearchSet.addAll(node.getStatementIds());
-            }
+                //check if the node's concept id is present in the alternative id list of the concept
```
jdamerow: pretty sure that's the wrong indentation

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
chiraag-subramanian: Have fixed the formatting issues and added the javadocs.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -88,20 +124,40 @@ public ITransformedNetwork getTransformedNetworkOfProject(String projectId, Stri
...
     }
-
+    
+    /**
+     * This method fetches ConceptPower entries related to the conceptId and return alternative id for the concept.
+     * 
+     * @param conceptId
+     * @return List<String>
+     */
+
```
jdamerow: remove empty line

jdamerow (approved):

jdamerow:
ok to test

jdamerow:
retest this please

jdamerow:
please remove the one empty line and it should be good to go

chiraag-subramanian:
Removed the empty lines.

PR #369, opened 2017-04-17
Story/quad 230

susmitha8:
Addressed the review comments and made changes accordingly.

### DIFF
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/rest/open/NodeInfoController.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/rest/open/NodeInfoController.java
@@ -19,6 +19,7 @@ import org.springframework.web.bind.annotation.RequestParam;
 
 import edu.asu.spring.quadriga.domain.enums.EProjectAccessibility;
 import edu.asu.spring.quadriga.domain.enums.ETextAccessibility;
+import edu.asu.spring.quadriga.domain.impl.projectblog.ProjectBlogEntry;
 import edu.asu.spring.quadriga.service.network.INetworkManager;
 import edu.asu.spring.quadriga.service.network.domain.impl.TextOccurance;
 import edu.asu.spring.quadriga.service.network.domain.impl.TextPhrase;
@@ -45,6 +46,9 @@ public class NodeInfoController {
                     && occur.getProject().getProjectAccess() == EProjectAccessibility.PUBLIC) {
                 JSONObject occurance = new JSONObject();
                 occurance.append("text", occur.getTextUri());
+                occurance.append("textAuthor", occur.getAuthor());
+                occurance.append("textTitle", occur.getTitle());
+                occurance.append("textCreationDate", occur.getCreationDate());
                 JSONArray phraseArray = new JSONArray();
 
                 List<TextPhrase> phrases = occur.getTextPhrases();
@@ -97,6 +101,9 @@ public class NodeInfoController {
                 if (occur.getProject().getProjectAccess() == EProjectAccessibility.PUBLIC) {
                     JSONObject occurance = new JSONObject();
                     occurance.append("text", occur.getTextUri());
+                    occurance.append("textAuthor", occur.getAuthor());
+                    occurance.append("textTitle", occur.getTitle());
+                    occurance.append("textCreationDate", occur.getCreationDate());
                     occurance.append("projectUnix", occur.getProject().getUnixName());
                     occurance.append("projectName", occur.getProject().getProjectName());
                     otherProjectsTexts.put(occurance);
--- a/Quadriga/src/main/java/edu/asu/spring/quadriga/rest/open/NodeInfoController.java
+++ b/Quadriga/src/main/java/edu/asu/spring/quadriga/rest/open/NodeInfoController.java
@@ -6,6 +6,7 @@ import java.util.List;
 import java.util.Set;
 
 import org.codehaus.jettison.json.JSONArray;
+import org.codehaus.jettison.json.JSONException;
 import org.codehaus.jettison.json.JSONObject;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.context.annotation.PropertySource;
@@ -13,14 +14,12 @@ import org.springframework.core.env.Environment;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Controller;
-import org.springframework.ui.Model;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
 import org.springframework.web.bind.annotation.RequestParam;
 
 import edu.asu.spring.quadriga.domain.enums.EProjectAccessibility;
 import edu.asu.spring.quadriga.domain.enums.ETextAccessibility;
-import edu.asu.spring.quadriga.domain.impl.projectblog.ProjectBlogEntry;
 import edu.asu.spring.quadriga.service.network.INetworkManager;
 import edu.asu.spring.quadriga.service.network.domain.impl.TextOccurance;
 import edu.asu.spring.quadriga.service.network.domain.impl.TextPhrase;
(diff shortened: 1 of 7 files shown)


Quadriga/src/main/java/edu/asu/spring/quadriga/rest/open/NodeInfoController.java
```diff
@@ -45,6 +46,7 @@
                     && occur.getProject().getProjectAccess() == EProjectAccessibility.PUBLIC) {
                 JSONObject occurance = new JSONObject();
                 occurance.append("text", occur.getTextUri());
```
jdamerow, line 48: this line seems to be the same in both cases too. Is there a reason this is not in the appendCall method?

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/open/NodeInfoController.java
```diff
@@ -111,4 +114,11 @@ public int compare(TextPhrase o1, TextPhrase o2) {
         return new ResponseEntity<String>(result.toString(), HttpStatus.OK);
 
     }
+
+    public void appendCall(JSONObject occurance, TextOccurance occur) throws JSONException {
```
jdamerow, line 118: bad method name. what does this method do?

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
```diff
@@ -15,13 +17,14 @@ function defineListeners(cy, path, unixName) {
 		var node = e.cyTarget;
 		conceptDescription(path, node);
 		getTexts(node, path, unixName);
+		getEntireText(node,path);
```
jdamerow, line 20: this would mean that when a user clicks on a node, the function getEntireText should be called. Which it shouldn't. The text should be shown when the user clicks on a link in the side bar, not when clicking on a node.

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
```diff
@@ -183,11 +189,21 @@ function getTexts(node, path, unixName) {
 				parsedData = JSON.parse(data);
 				var projects = parsedData['projects'];
 				projects.forEach(function(element, index, array) {
-					conceptDesc += "<p class='text-muted' style='margin-bottom:2px'>" + element["text"] + "</p>";
-					conceptDesc += "... " + hightlight(element["textContent"][0], element["phrases"]) + "..." + "<br>";
+					handler += "<a href="+ "'" + element["text"] + "'" + "target='_blank'>" + element["text"] + "</a>";
```
jdamerow, line 192: this would append the link for every project to the handler variable and then later on put that into the concept description. when there are two projects the second project would have the value from the first project as well as the value from the second project (given that you just append it).

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
```diff
@@ -168,6 +171,9 @@ function conceptDescription(path, node) {
 
 function getTexts(node, path, unixName) {
 	var conceptDesc = "";
+	var handler = "";
+	var sample = "";
```
jdamerow, line 175: what is this for?

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
```diff
@@ -183,11 +189,21 @@ function getTexts(node, path, unixName) {
 				parsedData = JSON.parse(data);
 				var projects = parsedData['projects'];
 				projects.forEach(function(element, index, array) {
-					conceptDesc += "<p class='text-muted' style='margin-bottom:2px'>" + element["text"] + "</p>";
-					conceptDesc += "... " + hightlight(element["textContent"][0], element["phrases"]) + "..." + "<br>";
+					handler += "<a href="+ "'" + element["text"] + "'" + "target='_blank'>" + element["text"] + "</a>";
+					conceptDesc += "<a href='#' data-toggle='modal' data-target='#txtModal' data-txtid="+ element["textId"] + " data-txttitle=" + element["textTitle"] + " data-txtauthor=" + element["textAuthor"] + " data-txtdate=" + element["textCreationDate"] + ">" + element["textAuthor"] + ", " + element["textTitle"] + ":" + "(" + element["textCreationDate"] + ")" + "</a>";
+					//conceptDesc += "<p class='text-muted' style='margin-bottom:2px'>" + element["textTitle"] + "</p>";
+					//conceptDesc += "<p class='text-muted' style='margin-bottom:2px'>" + element["textCreationDate"] + "</p>";
```
jdamerow, line 195: remove code you don't longer need

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
```diff
@@ -183,11 +189,21 @@ function getTexts(node, path, unixName) {
 				parsedData = JSON.parse(data);
 				var projects = parsedData['projects'];
 				projects.forEach(function(element, index, array) {
-					conceptDesc += "<p class='text-muted' style='margin-bottom:2px'>" + element["text"] + "</p>";
-					conceptDesc += "... " + hightlight(element["textContent"][0], element["phrases"]) + "..." + "<br>";
+					handler += "<a href="+ "'" + element["text"] + "'" + "target='_blank'>" + element["text"] + "</a>";
+					conceptDesc += "<a href='#' data-toggle='modal' data-target='#txtModal' data-txtid="+ element["textId"] + " data-txttitle=" + element["textTitle"] + " data-txtauthor=" + element["textAuthor"] + " data-txtdate=" + element["textCreationDate"] + ">" + element["textAuthor"] + ", " + element["textTitle"] + ":" + "(" + element["textCreationDate"] + ")" + "</a>";
```
jdamerow, line 193: same is true here. you're accumulating concept descriptions with every project.

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
```diff
@@ -197,6 +213,43 @@ function getTexts(node, path, unixName) {
 	});
 }
 
+function getEntireText(node,path){
+	$(document).ready(function() {
+	$('#txtModal').on('show.bs.modal',function(event){
```
jdamerow, line 218: either you have a function or you do something on document ready. this does not make sense.

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/open/NodeInfoController.java
susmitha8: I thought of adding this also in the method but in the earlier review request this wasn't highlighted to be added in the method, so haven't added that. Will add it.

Quadriga/src/main/java/edu/asu/spring/quadriga/rest/open/NodeInfoController.java
susmitha8: As this method appends JSONObject key with its value I named it that way. Should this be changed?

jdamerow: method names should say what they actually do (on a semantic level, not on a technical level), otherwise they are difficult to reuse if nobody knows what they are doing by looking at their name.

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
susmitha8: Variable "handler" is being used in the function later on but the variable "sample" is not being used anywhere. So will delete the variable "sample".

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
susmitha8: will remove it

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
susmitha8: Understood. Removed the "getEntireText" function call from defineListeners function.

Quadriga/src/main/webapp/resources/js/cytoscape/publicNetwork.js
susmitha8: True. Removed the document ready function.

PR #393, opened 2017-10-23
[QUAD-260] Removed JAXB annotations from DTOs.

chiraag-subramanian:


abhikrish503501 (approved):
The annotations have been removed properly and indentation is good

Mihika0409 (approved):

jdamerow:
ok to test

Quadriga/src/main/java/edu/asu/spring/quadriga/dto/WorkspaceDTO.java
```diff
@@ -21,7 +21,6 @@
 import javax.persistence.OneToMany;
 import javax.persistence.OneToOne;
 import javax.persistence.Table;
-import javax.xml.bind.annotation.XmlTransient;
```
jdamerow, line 24: don't delete just the imports; you also need to delete where they are used. eclipse should have given you an error if you do that.

jdamerow (changes requested):
make sure you don't break the code; eclipse should have been all over with compiler errors;

abhikrish503501 (approved):
Removed the annotations

jdamerow (approved):

PR #395, opened 2018-02-21
Story/quad 264

chiraag-subramanian:


Quadriga/src/main/webapp/WEB-INF/views/public/searchednetwork.jsp
```diff
@@ -0,0 +1,214 @@
...
+    document.addEventListener('fullscreenchange', exitHandler, false);
+    document.addEventListener('MSFullscreenChange', exitHandler, false);
+}
+
+function exitHandler()
+{
+    if (document.webkitIsFullScreen || document.mozFullScreen || document.msFullscreenElement !== null)
+    {
+    	if(window.innerWidth == screen.width && window.innerHeight == screen.height) {
+    	}
```
Mihika0409: The else can probably be removed by modifying the if condition

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +93,53 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
 
         return sb.toString();
     }
+    
+    @Override
+    public List<CytoscapeNodeObject> getNodes(Map<String, Node> nodes){
+        List<CytoscapeNodeObject> cytoscapeNodeList = new ArrayList<CytoscapeNodeObject>();
+        List<Node> nodeList = new ArrayList<Node>(nodes.values());
+        for (int i = 0; i < nodeList.size(); i++) {
```
jdamerow: why don't you use a foreach loop ```for (Node node : nodeList)``` and get rid of the all get() calls? would make your code much more readable.

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/network/json/PublicSearchObject.java
```diff
@@ -0,0 +1,44 @@
...
+
+    public List<CytoscapeLinkObject> getLinks() {
+        return links;
+    }
+
+    public void setLinks(List<CytoscapeLinkObject> links) {
+        this.links = links;
+    }
+
+}
```
jdamerow: this class and all the other cytoscape classes seem to be purely for the presentation layer and shoudln't be in the domain package.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +93,53 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+                dataObj.setGroup("1");
+            }
+            dataObj.setSourceReference(nodeList.get(i).getSourceReference());
+            dataObj.setStatementIds(new ArrayList<String>());
+            for (int j = 0; j < nodeList.get(i).getStatementIds().size(); j++) {
+                dataObj.getStatementIds().add(nodeList.get(i).getStatementIds().get(j)); 
+            }
+            
+            CytoscapeNodeObject obj = new CytoscapeNodeObject();
+            obj.setData(dataObj);
```
jdamerow: creating a node object should be extracted into a private method

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +93,53 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+            CytoscapeLinkDataObject dataObj = new CytoscapeLinkDataObject();
+            dataObj.setId(""+i);
+            dataObj.setSource(links.get(i).getSubject().getId());
+            dataObj.setTarget(links.get(i).getObject().getId()); 
+            dataObj.setLabel(links.get(i).getLabel());
+            dataObj.setSourceReference(links.get(i).getSourceReference());
+            dataObj.setStatementIds(new ArrayList<String>());
+            dataObj.getStatementIds().add(links.get(i).getStatementId()) ;
+            CytoscapeLinkObject obj = new CytoscapeLinkObject();
+            obj.setData(dataObj);
```
jdamerow: creating of links should be in its own private method

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
```diff
@@ -63,6 +72,10 @@
 
     private static final Logger logger = LoggerFactory.getLogger(NetworkSearchController.class);
 
+    private static ExecutorService executorService = Executors.newFixedThreadPool(10);
```
jdamerow: executorservice should be configured in either xml or config bean. also, why is this static?

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
```diff
@@ -63,6 +72,10 @@
...
+
+    private static Map<Integer, Future<PublicSearchObject>> searchResultMap = new HashMap<Integer, Future<PublicSearchObject>>();
```
jdamerow: why is this static?

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
```diff
@@ -131,53 +146,104 @@ public String getSearch(@ProjectIdentifier @PathVariable("projectUnixName") Stri
...
-        if (transformedNetwork != null) {
-            json = jsonCreator.getJson(transformedNetwork.getNodes(), transformedNetwork.getLinks());
-        }
+        Callable<PublicSearchObject> callable = () -> {
 
-        if (transformedNetwork == null || transformedNetwork.getNodes().size() == 0) {
-            model.addAttribute("isNetworkEmpty", true);
+            PublicSearchObject publicSearchObject = new PublicSearchObject();
+            ITransformedNetwork transformedNetwork = transformationManager
+                    .getSearchTransformedNetwork(project.getProjectId(), conceptId, INetworkStatus.APPROVED);
```
jdamerow: ```getSearchTransformedNetwork``` should be called as a asycn method (```@async``` annotation). turning the resulting network into json should be done in the method that checks for the results.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
```diff
@@ -131,53 +146,104 @@ public String getSearch(@ProjectIdentifier @PathVariable("projectUnixName") Stri
...
+        Random randomTokenGenerator = new Random();
+        Integer randomToken = randomTokenGenerator.nextInt(100);
+        while (searchResultMap.containsKey(randomToken)) {
+            randomToken = randomTokenGenerator.nextInt(100);
         }
-
-        model.addAttribute("jsonstring", json);
-        model.addAttribute("networkid", "\"\"");
-        model.addAttribute("project", project);
+        searchResultMap.put(randomToken, future);
```
jdamerow: there should be a service that keeps track of running searches. This is not the job of the controller.

jdamerow: also, your from the token that you use to check on the progress the original search string should be retrievable. you don't want to start the same search again, if a user searches for the same query again shortly after it was submitted.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
```diff
@@ -131,53 +146,104 @@ public String getSearch(@ProjectIdentifier @PathVariable("projectUnixName") Stri
...
+            publicSearchObject.setStatus(2);
+            return publicSearchObject;
+        }
+
+        Future<PublicSearchObject> futureResult = searchResultMap.get(tokenId);
+        if (futureResult != null && futureResult.isDone()) {
+            try {
+                publicSearchObject = futureResult.get();
+                publicSearchObject.setStatus(1);
+                searchResultMap.remove(tokenId);
```
jdamerow: also, in the service that manages pending searches I would use a cache that expires the searches at some point but will keep them around for a while (e.g. if the user reloads the page).

Quadriga/src/main/webapp/WEB-INF/views/public/searchednetwork.jsp
```diff
@@ -0,0 +1,215 @@
...
+
+<script
+	src="${pageContext.servletContext.contextPath}/resources/js/jquery-1.9.1.min.js"></script>
+<script
+	src="${pageContext.servletContext.contextPath}/resources/js/cytoscape/dist/cytoscape.js"></script>
+
+
+<div class="row">
+	<div class="alert alert-info" id="load">Searching for network
+		containing the term ${searchNodeLabel}...</div>
```
jdamerow: Should be simply "Searching for ${searchNodeLabel}. This might take a little while, please wait..."

Quadriga/src/main/webapp/WEB-INF/views/public/searchednetwork.jsp
```diff
@@ -0,0 +1,215 @@
...
+	<div class="alert alert-info" id="network-empty" style="display: none;">Could
+		not find network containing the term ${searchNodeLabel}</div>
```
jdamerow: Should be "Could not find any statements containing ${searchNodeLabel}."

Quadriga/src/main/webapp/WEB-INF/views/public/searchednetwork.jsp
```diff
@@ -0,0 +1,215 @@
...
+	<div class="alert alert-info" id="network-error" style="display: none;">Server
+		could not generate a response...Try again later.</div>
```
jdamerow: Sorry, there was a server error. Please try again later.

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/domain/network/json/PublicSearchObject.java
chiraag-subramanian: Should I create a new package under quadriga called presentation...or is there an existing package where I can place these presentation layer classes?

jdamerow: they should be in a subpackage of the package where the controllers are in

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -3,24 +3,59 @@
...
-    /*
-     * (non-Javadoc)
-     * 
-     * @see
-     * edu.asu.spring.quadriga.service.network.impl.JsonCreator#getJson(java.
-     * util.Map, java.util.List)
-     */
+    //private ExecutorService executorService = Executors.newFixedThreadPool(10);
+    //@Autowired
+    //private ExecutorService executorService;
```
jdamerow: removed commented out code

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -3,24 +3,59 @@
...
+    @Autowired
+    private ThreadPoolTaskExecutor taskExecutor;
+    private static final Logger logger = LoggerFactory.getLogger(CytoscapeJsonCreator.class);
```
jdamerow: doesn't need to be static

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +124,108 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+            dataObj.setConceptUri(node.getConceptId());
+            dataObj.setConceptId(node.getConceptIdShort());
+            if(node instanceof PredicateNode){
+                dataObj.setGroup("0");
+            }
+            else{
+                dataObj.setGroup("1");
+            }
+            dataObj.setSourceReference(node.getSourceReference());
+            dataObj.setStatementIds(node.getStatementIds());
```
jdamerow: the whole creation of the object should be in the create method

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +124,108 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+        for(Link link : links){
+            CytoscapeLinkDataObject dataObj = new CytoscapeLinkDataObject();
+            dataObj.setId(""+i);
+            dataObj.setSource(link.getSubject().getId());
+            dataObj.setTarget(link.getObject().getId());
+            dataObj.setLabel(link.getLabel());
+            dataObj.setSourceReference(link.getSourceReference());
+            dataObj.setStatementIds(new ArrayList<String>());
+            dataObj.getStatementIds().add(link.getStatementId());
+            cytoscapeLinkList.add(createCytoscapeLinkObject(dataObj));
```
jdamerow: same as above

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +124,108 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+                    .getSearchTransformedNetwork(project.getProjectId(), conceptId, INetworkStatus.APPROVED);
+                  if (transformedNetwork != null && transformedNetwork.getNodes().size() > 0) {
+                      publicSearchObject.setNodes(getNodes(new ArrayList<Node>(transformedNetwork.getNodes().values())));
+                      publicSearchObject.setLinks(getLinks(transformedNetwork.getLinks()));
+                  }
+                  else{
+                      publicSearchObject.setNetworkEmpty(true);
+                  }
+                  return publicSearchObject;
+              };
```
jdamerow: this should be in its own class. there should be one class per task (token management, performing a long running task, creating json, etc) -> SRP (single responsibility principle) -> a class should have only one reason to change.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +124,108 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+        }
+        return conceptIdToken;
+    }
+    
+    @Override
+    public PublicSearchObject getSearchTransformedNetwork(String conceptIdToken){
+        
+        PublicSearchObject publicSearchObject = new PublicSearchObject();
+        if(!cache.isKeyInCache(conceptIdToken)){
+            publicSearchObject.setStatus(2);
```
jdamerow: what is status 2?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +124,108 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+        if (futureResult != null && futureResult.isDone()) {
+            try {
+                publicSearchObject = futureResult.get();
+                publicSearchObject.setStatus(1);
+            } catch (InterruptedException | ExecutionException e) {
+                logger.error("Exception while retrieving the result", e);
+                publicSearchObject.setStatus(2);
+            }
+        } else {
+            publicSearchObject.setStatus(3);
```
jdamerow: integers are almost always bad status indicator? what does 1, 2, or 3 mean? so hard to maintain.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/cytoscapeobjects/CytoscapeLinkDataObject.java
```diff
@@ -0,0 +1,62 @@
+package edu.asu.spring.quadriga.web.publicwebsite.cytoscapeobjects;
```
jdamerow: bad package name. 'publicwebsite' is already not very good, but adding 'cytoscapeobjects' to it makes it even worse. Find a short name that's to the point.

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/cytoscapeobjects/PublicSearchObject.java
```diff
@@ -0,0 +1,44 @@
+package edu.asu.spring.quadriga.web.publicwebsite.cytoscapeobjects;
+
+import java.util.List;
+
+public class PublicSearchObject {
+
+    boolean isNetworkEmpty;
+    List<CytoscapeNodeObject> nodes;
+    List<CytoscapeLinkObject> links;
+    int status;
```
jdamerow: why are they not private? are they needed anywhere else?

Quadriga/src/main/webapp/WEB-INF/views/public/searchednetwork.jsp
```diff
@@ -0,0 +1,212 @@
...
+function exitHandler()
+{
+    if (document.webkitIsFullScreen || document.mozFullScreen || document.msFullscreenElement !== null)
+    {
+    	if(!(window.innerWidth == screen.width && window.innerHeight == screen.height)) {
+    		clear();
+    	}
+    }
+}
+</script>
```
jdamerow: why are there so many script blocks?

jdamerow (changes requested):
javadoc is missing for several classes

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
chiraag-subramanian: Can I use an Enum while setting the status...to improve the code readability?

jdamerow: enums are always a good idea if you have only a limited number of choices :)

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
```diff
@@ -0,0 +1,95 @@
...
+    private final Logger logger = LoggerFactory.getLogger(AsyncNetworkTransformationService.class);
+    
+    
+    @PostConstruct
+    public void init() {
+        cache = cacheManager.getCache("asyncTransformedNetworks");
+    }
+    
+    
+    private String generateToken(String conceptId){
```
jdamerow, line 53: if you have to cut off the last part of the uri to get an id, then it's not an id but a uri. your variable name should reflect that.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
```diff
@@ -0,0 +1,95 @@
...
+        return conceptId.substring(conceptId.lastIndexOf('/') + 1);
+    }
+    
+    
+    @Override
+    public String submitNetworkTransformationRequest(String conceptId, IProject project){
+        String token = generateToken(conceptId);
+        if(!cache.isKeyInCache(token)){
```
jdamerow, line 61: from the javadoc:
"Since no assertions are made about the state of the Element it is possible that the Element is expired, but this method still returns true."
in that case, you want to restart the transformation, don't you?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
```diff
@@ -0,0 +1,95 @@
...
+             cache.put(new Element(token, transformationManager.getTransformedNetwork(project.getProjectId(), conceptId, INetworkStatus.APPROVED)));
+        }
+        return token;
+    }
+    
+    @Override
+    public TransformationRequestStatus getTransformationRequestStatus(String token){
+        if(!cache.isKeyInCache(token)){
+             return TransformationRequestStatus.FAILED;
```
jdamerow, line 70: that seems like the wrong result code to return. if the token doesn't exist in the cache, the process didn't necessarily fail, did it? but rather someone tries to retrieve a network, without having started the transformation?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+        Future<ITransformedNetwork> futureResult = (Future<ITransformedNetwork>) cache.get(token).getObjectValue();
```
jdamerow, line 72: what if the element is expired?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
```diff
@@ -0,0 +1,95 @@
...
+        
+        return TransformationRequestStatus.COMPLETE;
+    }
+    
+    @Override
+    public ITransformedNetwork getTransformedNetwork(String token){
+        if(getTransformationRequestStatus(token).equals(TransformationRequestStatus.COMPLETE)){
+            try {
+                    Future<ITransformedNetwork> futureResult = (Future<ITransformedNetwork>) cache.get(token).getObjectValue();
+                    return futureResult.get();
```
jdamerow, line 86: formatting

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
```diff
@@ -0,0 +1,95 @@
...
+            } catch (InterruptedException | ExecutionException e) {
+                logger.error("Exception while retrieving the result", e);
+                cache.remove(token);
+                return null;
```
jdamerow, line 90: what happens in this case? from the code, I guess that whoever is calling this method, could call it until they get a result. But if in the case that something went wrong it returns the same value as when the process isn't done yet, this would cause an infinite calling cycle.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+        return null;
+    }
+}
```
jdamerow, line 95: this class needs really good javadoc, on the class level as well as method level, that explains how to use it.

jdamerow: this class also needs test cases

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +92,97 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+     * @author Chiraag Subramanian
+     */
+    @Override
+    public List<CytoscapeLinkObject> getLinks(List<Link> links){
+
+        List<CytoscapeLinkObject> cytoscapeLinkList = new ArrayList<CytoscapeLinkObject>();
+        int i = 0;
+        for(Link link : links){
+            cytoscapeLinkList.add(createCytoscapeLinkObject(link));
+            cytoscapeLinkList.get(i).getData().setId(""+i);
```
jdamerow, line 125: setting the id should be done in the create method

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +92,97 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+        dataObj.setStatementIds(new ArrayList<String>());
+        dataObj.getStatementIds().add(link.getStatementId());
+
+        CytoscapeLinkObject obj = new CytoscapeLinkObject();
+        obj.setData(dataObj);
+        return obj;
+    }
+    
+    /**
+     * This method constructs the response object consisting of the transformed network and the status of the async network transformation.
```
jdamerow, line 168: what if the transformation isn't done yet?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +92,97 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+     * @author Chiraag Subramanian
+     */
+    @Override
+    public CytoscapeSearchObject getCytoscapeSearchObject(ITransformedNetwork transformedNetwork, TransformationRequestStatus transformationRequestStatus) {
+        CytoscapeSearchObject cytoscapeSearchObject = new CytoscapeSearchObject();
+        if (transformedNetwork != null && transformedNetwork.getNodes().size() > 0) {
+            cytoscapeSearchObject.setNodes(getNodes(new ArrayList<Node>(transformedNetwork.getNodes().values())));
+            cytoscapeSearchObject.setLinks(getLinks(transformedNetwork.getLinks()));
+        }
+        else{
```
jdamerow, line 181: please follow our style guide

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +92,97 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+        CytoscapeLinkObject obj = new CytoscapeLinkObject();
+        obj.setData(dataObj);
+        return obj;
+    }
+    
+    /**
+     * This method constructs the response object consisting of the transformed network and the status of the async network transformation.
+     * @param  transformedNetwork
+     * @param  transformationRequestStatus
+     * @return CytoscapeSearchObject
```
jdamerow, line 171: parameters need descriptions

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/NetworkTransformationManager.java
```diff
@@ -351,4 +356,21 @@ private ITransformedNetwork getFinalTransformedNetwork(ITransformedNetwork trans
...
+     * @return Future<ITransformedNetwork>
+     * @author Chiraag Subramanian
+     */
+    @Override
+    public Future<ITransformedNetwork> getTransformedNetwork(String projectId, String conceptId, String status){
+        Callable<ITransformedNetwork> callable = () -> {
+            return getSearchTransformedNetwork(projectId, conceptId, status);
+        };
+        
+        return taskExecutor.submit(callable);
```
jdamerow, line 374: just annotate getSearchTransformedNetwork with Async and get rid of this method

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/NetworkSearchController.java
```diff
@@ -131,53 +133,61 @@ public String getSearch(@ProjectIdentifier @PathVariable("projectUnixName") Stri
...
-     * given projects that contain the provided concept id.
-     * @param projectUnixName Unix name of the project whose networks will be searched for the concept
-     * @param conceptId Id of the concept
+     * This method accepts a request to fetch network corresponding to the
+     * searched term. The task of generating the network is delegated to a
+     * separate thread and a token (which identifies the submitted request) is
+     * generated.
+     * 
+     * @param projectUnixName
+     * @param conceptId
```
jdamerow, line 142: what happened to the parameter descriptions?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/CytoscapeJsonCreator.java
```diff
@@ -89,4 +92,97 @@ public String getJson(Map<String, Node> nodes, List<Link> links) {
...
+            cytoscapeSearchObject.setLinks(getLinks(transformedNetwork.getLinks()));
+        }
+        else{
+            cytoscapeSearchObject.setNetworkEmpty(true);
+        }
+        cytoscapeSearchObject.setStatus(transformationRequestStatus.getStatusCode());
+        
+        return cytoscapeSearchObject;
+    }
+}
```
jdamerow, line 188: needs test cases

Quadriga/src/main/java/edu/asu/spring/quadriga/web/publicwebsite/graph/CytoscapeLinkDataObject.java
```diff
@@ -0,0 +1,62 @@
+package edu.asu.spring.quadriga.web.publicwebsite.graph;
+
+import java.util.List;
+
+public class CytoscapeLinkDataObject {
```
jdamerow, line 5: all these classes need javadoc

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
```diff
@@ -0,0 +1,95 @@
...
+import edu.asu.spring.quadriga.service.network.IAsyncNetworkTransformationService;
+import edu.asu.spring.quadriga.service.network.domain.ITransformedNetwork;
+import edu.asu.spring.quadriga.web.network.INetworkStatus;
+import net.sf.ehcache.Cache;
+import net.sf.ehcache.CacheManager;
+import net.sf.ehcache.Element;
+
+/**
+ * This class is responsible for initiating and managing async transformation requests.
+ * It uses LRU cache to maintain the results for an hour.
```
jdamerow, line 26: how long results are cached is configurable, so it might not be an hour.

jdamerow: same with the type of cache

jdamerow (changes requested):

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
chiraag-subramanian: No. When the server encounters an error, it removes the token from the cache. In the subsequent request from the client, the token won't exist in the cache. So the server will inform the client that the token doesn't exist: indicating the server faced an error or the client request is not valid. The client stops making the request and displays the appropriate message.

jdamerow: if I only use this method and continuously check for the result of this method, I will not know that something failed, as the method always returns null (if I never call getTransformationRequestStatus). Ideally you would handle failure and not done yet differently in this method, or you need to be very clear and specific in the javadoc for this class. Intuitively, nothing tells me that I can't just call this method until i get a result.

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
chiraag-subramanian: Should I call it INVALID instead of FAILED ?

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
chiraag-subramanian: I don't think that the cache entry has expired. The documentation says that if the cache entry expires, it is automatically removed from the cache.
http://www.ehcache.org/documentation/3.1/caching-concepts.html

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
chiraag-subramanian: I don't think that we need to check if the cache entry has expired. The documentation says that if the cache entry expires, it is automatically removed from the cache.
http://www.ehcache.org/documentation/3.1/caching-concepts.html

Quadriga/src/main/java/edu/asu/spring/quadriga/service/network/impl/AsyncNetworkTransformationService.java
jdamerow: sure, there also might be the case though that the transformation actually failed (e.g. if an exception was thrown)?

jdamerow:
We’re currently using version 2.10, which according to http://www.ehcache.org/generated/2.10.0/html/ehc-all/#page/Ehcache_Documentation_Set%2Fco-life_options_affecting_data_life.html%23wwconnect_header <http://www.ehcache.org/generated/2.10.0/html/ehc-all/#page/Ehcache_Documentation_Set/co-life_options_affecting_data_life.html#wwconnect_header> does not remove necessarily remove expired elements right away.

---
Julia Damerow, PhD
Scientific Software Engineer
ASU-SFI Center for Biosocial Complex Systems
Digital Innovation Group
Arizona State University
email: jdamerow@asu.edu <mailto:jdamerow@asu.edu> <>
web: http://diging.asu.edu/ <http://diging.asu.edu/> <>

PR #396, opened 2018-04-26
Story/quad 271

chiraag-subramanian:
Functionality to enable admin users to add themselves as project admins to other projects.


jdamerow:
ok to test

